PR #9, opened 2015-09-05
Small enhancements to circle

jsanchezv:
ARM have two instructions for change the byte order in one instruction for 16 and 32 bits sizes.

Add support for virtual resolutions and depths < 8 bits.

### DIFF
--- a/include/circle/bcmframebuffer.h
+++ b/include/circle/bcmframebuffer.h
@@ -3,7 +3,7 @@
 //
 // Circle - A C++ bare metal environment for Raspberry Pi
 // Copyright (C) 2014  R. Stange <rsta2@o2online.de>
-// 
+//
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
 // the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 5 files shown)


rsta2:
Thanks for this! Because I was unsure what to do with it I did put it away from day to day and did not make a decision so far. As I feel forced to a decision now I have to tell I will unfortunately not include these changes into Circle. Sorry for that!

Let me explain why:

* Circle is in general not optimized for speed very much. I think the rewrite of le2be16() and le2be32() will not change this very much. The current implementation is translated into a few machine instructions only which should be OK.
* The change in frame buffer support cannot be tested with the standard Circle environment. I decided that only features will come into this repository which are tested by myself. So this would require additional effort for testing code I do not have time for at the moment.
* I do not want to change the repository if there isn't a sufficient gain from it because it causes additional effort for users of Circle to check if these changes are important for their own copy of Circle.

jsanchezv:
No problem man! :)

I want to write a bare-metal emulator, so the speed is a matter for me. Well, this code isn't the problem, of course, but I guess that less code == less bugs. Really, the ASM function ins't needed because a GCC builtin can be used:

int32_t __builtin_bswap32 (int32_t x)

and don't jump/return to a subroutine.

For the framebuffer code, I have a test derived from sample #02 that uses the new code. But two things needs to be know:

1.- When the Bcm2835FrameBufferInfo struct is created, reserves a fixed 256 palette entries. That's not a problem, anyway.
2.- The text drawing classes aren't ready to work with a depth != 8, I guess. That's not a problem for me, by the moment.

Anyway, thanks by your answer.

rsta2:
I revised my decision what is a "sufficient gain" for Circle because now I understand what you want to do with these changes. An emulator is an interesting application for Circle.

I didn't knew that the frame buffer can be used with depths below 8. I will merge the frame buffer changes untested because I suppose you have tested this sufficiently. Thanks for providing this enhancement.

Also thanks for the hint using __builtin_bswap{16,32}() instead of an own implementation of le2be{16,32}(). I will change *include/circle/util.h* like this:
```
#ifdef __GNUC__

#define le2be16		__builtin_bswap16

#define le2be32		__builtin_bswap32

#else

u16 le2be16 (u16 usValue);

u32 le2be32 (u32 ulValue);

#endif
```

rsta2:
I merged the changes to *CBcmFrameBuffer* manually because there were conflicts in *util.h*, *util.cpp* and *util_fast.S*.

PR #21, opened 2016-09-06
Fix another new[]/delete mismatch.

smuehlst:
During my experiments with circle I found another new[]/delete mismatch in the CLogger class.

BTW, wouldn't it make sense to replace the allocated CLogger::m_pBuffer pointer with a
member array of size LOGGER_BUFSIZE, as the size is known at compile time anyway?
This would avoid the new/delete for CLogger::m_pBuffer altogether.

### DIFF
--- a/lib/logger.cpp
+++ b/lib/logger.cpp
@@ -47,7 +47,7 @@ CLogger::~CLogger ()
 {
 	s_pThis = 0;
 
-	delete m_pBuffer;
+	delete [] m_pBuffer;
 	m_pBuffer = 0;
 
 	m_pTarget = 0;


rsta2:
Thanks for this pull request. I have merged it.

I think it's better to allocate such buffers on the heap which is much bigger than the kernel stack (only 128 KB) where the CLogger object is normally created. Because this object is constructed only once, while the system is initialized, there would be no performance gain.

smuehlst:
@rsta2 : I see, I wasn't aware of the limitation of the kernel stack. Can the size of the kernel stack be configured somehow?

rsta2:
Have a look at *include/circle/sysconfig.h*. The memory layout of the data regions behind the loaded kernel image is defined there.

You can increase the KERNEL_STACK_SIZE there. But be sure that it is a multiple of 16KB (so that MEM_PAGE_TABLE1 stays 16KB aligned) and that you don't run into MEM_HEAP_START and MEM_COHERENT_REGION (on RPi 3 only, must be on a 1MB boundary) which are set to a fixed address. Please note that on the RPi 2 and 3 there are four separate kernel stacks (one for each CPU core) of the defined size.

The following changes set the kernel stack size to 1MB:

```
#define KERNEL_STACK_SIZE	0x100000
...
#if RASPPI == 3
#define MEM_COHERENT_REGION	0x1000000
#define MEM_HEAP_START		0x1100000
#else
#define MEM_HEAP_START		0x1000000
#endif
```

PR #25, opened 2016-10-30
git-ignore all downloaded files in boot directory

smuehlst:
As the Raspberry Pi firmware will never be added to the circle repository, I added a corresponding .gitignore file to get a clean unmodified tree even with the downloaded firmware files being present.


rsta2:
Thank you!

PR #58, opened 2017-12-18
Eclipse support

Matheus-Garbelini:
Hello, in order to make things easier, and contribute to this repository so more people can make amazing things with this blazing fast framework with raspberry pi, i did the following:

- Modified main Rules.mk to include flash, monitor, all options so integration with eclipse is wellcome

- In order to make flash, I included a new folder called tools, including a python serial script and the source code to the modified bootloader (Blazing fast, no need to do the SD card dance)

- Included make monitor defaulting to putty, as it is works in linux and windows. Also it supports different baud rates.

- Included in the beginning of Rules.mk baudrate and serial port config to the flasher (this is identical to what espressif idf does so i think there's no problem in doing that).

- Modified makefile of boot folder, so it also compiles bootloader for raspberry1,2,3 (kernel.img and kernel7.img), thus allowing users to a faster getting started.

Thats it for now. Thanks for your support, feel free to edit anything, that's the first serious merge that i'm doing so i'm always open to learn.

I'm planning to allow this to be used by arduino users, opening the possibility to raspberrys be used even in robots competition (they know arduino and windows only xD).
As a electronic engineering student that like robotics and so on, it is very hard to me and other people from hardware area to understand what's going on in low level arm devices. So much that most of us can't install dependencies, nor use make, nor do things (in linux in general) that for a computer engineer's perspective is normal (at least here in Brazil).

Best regards.

### DIFF
--- a/boot/Makefile
+++ b/boot/Makefile
@@ -10,6 +10,10 @@ all: clean
 	wget -q -O bootcode.bin https://github.com/raspberrypi/firmware/blob/master/boot/bootcode.bin?raw=true
 	wget -q -O fixup.dat https://github.com/raspberrypi/firmware/blob/master/boot/fixup.dat?raw=true
 	wget -q -O start.elf https://github.com/raspberrypi/firmware/blob/master/boot/start.elf?raw=true
+	make -C "../tools/bootloader"
+	cp ../tools/bootloader/kernel.img .
+	cp ../tools/bootloader/kernel7.img .
+	make clean -C "../tools/bootloader"
 
 clean:
-	rm -f bootcode.bin fixup.dat start.elf LICENCE.broadcom
+	rm -f bootcode.bin fixup.dat start.elf kernel.img kernel7.img LICENCE.broadcom


rsta2:
Thanks for this contribution. I need some time to evaluate it.

Matheus-Garbelini:
@rsta2 thank you, i think the all: entry in Rules.mk can be removed

Rules.mk
```diff
@@ -91,7 +91,21 @@ $(TARGET).img: $(OBJS) $(LIBS) $(CIRCLEHOME)/lib/startup.o $(CIRCLEHOME)/circle.
 	$(LD) -o $(TARGET).elf -Map $(TARGET).map -T $(CIRCLEHOME)/circle.ld $(CIRCLEHOME)/lib/startup.o $(OBJS) $(EXTRALIBS) $(LIBS) $(EXTRALIBS)
 	$(PREFIX)objdump -d $(TARGET).elf | $(PREFIX)c++filt > $(TARGET).lst
 	$(PREFIX)objcopy $(TARGET).elf -O binary $(TARGET).img
+	$(PREFIX)objcopy $(TARGET).elf -O ihex $(TARGET).hex
 	wc -c $(TARGET).img
 
 clean:
 	rm -f *.o *.a *.elf *.lst *.img *.cir *.map *~ $(EXTRACLEAN)
+
+	flash:
```
MaPePeR: This looks wrong to me.

Matheus-Garbelini: yes, extra indentation, sorry.

rsta2:
Thanks again for this valuable contribution. I would take these features under the headline "Eclipse support" and have added a feature branch *eclipse-support* to the Circle repository, because it think, there is still some work to do, which can be done on this branch, before merging it to *develop*.

Starting with your pull request, I did some changes to several files. Please have a look at the commit notices for details. There are two main changes in behaviour from your version:

* Serial baud rate is configurable and is 115200 by default now. I think, the value of 1008065 is not a standard and not supported by each serial interface. But you can use the baud rate of your choice, if you create a file *Config.mk* in Circle home with the line `DEFAULTBAUD = 1008065`. The bootloader will automatically be build for this baud rate then. This does not work yet with *tools/flasher.py* at the moment (see below).

* Using `make` in the boot/ directory only downloads the firmware files as before, because probably there will be users, who do not need the bootloader and are used to find the firmware files only in boot/. But you can enter `make all` in boot/ to download the firmware files and build the bootloader too.

There is still some work to do on the *flasher.py* tool. It would be great, if you could do that, because I'm not a python programmer. It must be made configurable for the serial port and baud rate (already given as parameter in *Rules.mk*) and should get some error handling (e.g. serial device or ihex input file cannot be opened).

I have tested the bootloader on RPi 2, 3 and Zero with a terminal program. I did not test *tools/flasher.py* and the whole feature with Eclipse so far.

Matheus-Garbelini:
@rsta2 tahnk you a lot. I will work this things and update you soon. the eclipse-suport branch is a great idea. In the future we may have an arduino-support also.

rsta2:
I tested this new feature as whole together with Eclipse now on Linux for the first time. Unfortunately I have problems with it. I created a build target "flash" and *flasher.py* can be run successfully from it, but I was not able to run a Circle sample successfully using the boot loader with it. When I start putty after "flashing" I have to press the 'g' key to get any reaction. The boot loader responds with "--" then, but I cannot get further serial output from the sample.

Because I want to prepare a new Circle release on the master branch soon, I will probably not include the Eclipse support into this release, because of this problem. I would also need a documentation file, which describes how to use Eclipse to develop Circle applications, which does not yet exist, and have to do some additional minor changes.

Matheus-Garbelini:
@rsta2 which version of raspberry pi are you using?

rsta2:
I was using a RPi 2 (with BCM2836). Yes, I can see the IHEX message.

rsta2:
I did some tests with *flasher.py* now and I think I've found a fix for my problems with it. I inserted a `sleep(1)` after sending the IHEX file and before sending the 'g' letter. With this change it is working well. The updated *flasher.py* is on the *eclipse-support* feature branch. There was also a problem with the *Makefile* of the bootloader, which has been fixed.

Matheus-Garbelini:
@rsta2 thank you. Good to know that was just a timing issue with python.
I Didn't have time to test it with raspberry pi 2 yet. Did you test it also with pi zero?

rsta2:
I tested it with RPi 1 (Model A+), 2, 3 and RPi Zero. All is running well.

PR #63, opened 2018-01-03
Eclipse support: Flasher.py update

Matheus-Garbelini:
->Support for error handling
->Support for different flash baudrate

Rules.mk update
->FLASHBAUD now defines bootloader serial baudrate instead of DEFAULTBAUD
which is used only for normal serial communication when doing make monitor

bootloader/Makefile.mk update
->changed DEFAULTBAUD to FLASHBAUD as above

### DIFF
--- a/Rules.mk
+++ b/Rules.mk
@@ -102,12 +102,13 @@ clean:
 
 SERIALPORT  ?= /dev/ttyUSB0
 DEFAULTBAUD ?= 115200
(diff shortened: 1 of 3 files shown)


rsta2:
Thanks for updating Flasher.py! I have merged it. I still have to test it all with Eclipse together with the release tests for the next Circle release.

PR #70, opened 2018-04-18
Alpha GDB Stubs

Julio-Guerra:
Hello,

A user asked for debugging Circle with our GDB stubs (https://github.com/farjump/raspberry-pi). There was almost nothing else to do other than using your existing `USE_RPI_STUB_AT` macro, but I couldn't properly provide the cache-coherent memory region to Circle.

So by changing current assumptions on cache-coherency and by making them explicit everywhere in Circle, I could make it work with a memory mapping other than the default one (we chose a MMU configuration working for every RPi model). It also has the benefit of clearly stating what memory access properties Circle requires: one with cache-coherency, one uncached, etc.

To do so, I basically modified `GetCoherentPage()` to no longer assume `0x0...` is cache-coherent, and introduced `GetUncachedAlias()` to get the uncached alias of a memory pointer. I then replaced every `GPU_MEM_BASE + ...` with a call to it. Note that they should be reviewed as I wasn't every time sure of whether or not getting the uncached alias was really intended.

So changing the MMU configuration is a matter of setting the different macros. For example,
to debug Circle with our gdb stubs:
```
CFLAGS += -DUSE_RPI_STUB_AT=0x7F00000 -DGPU_IO_BASE=0x80000000 \
          -DGPU_UNCACHED_BASE=0x40000000 -DGPU_CACHE_COHERENT_BASE=0x40000000 \
          -DARM_IO_BASE=0x20000000
OPTIMIZE := -O0 -g3 -ggdb
```

This PR is mostly to start discussing. We could definitely go much further and make debugging a first-class feature of Circle. Among our best features: GDB multi-core debugging, GDB semihosting (for example to print Circle logs in GDB's console), and GDB's stop signal.


rsta2:
Hello, to be able to debug Circle applications using Alpha can be very interesting. rpi_stub, which has been supported until now, has not been maintained for some time and I think, Alpha has more features. I'll need some time to evaluate this.

rsta2:
I have created a new feature branch *alpha-gdbstubs-support* in the Circle repository and pushed some commits to it. I used your pull request for information but implemented the Alpha GDB server support slightly different after analysing Alpha's MMU translation table. I have tested it successfully on RPi A+, Zero and 3B.

Some things to note:

* I have added a Circle system option *USE_ALPHA_STUB_AT*, which can be defined in *include/circle/sysconfig.h* to use this support. Alternatively you can define in *Config.mk*:


```
USE_ALPHA_STUB_AT = 0x7F00000
OPTIMIZE = -O0 -g3 -ggdb
```

* The macro *BUS_ADDRESS*, which is defined in *include/circle/bcm2835.h* does most of the job. Please note, that GPU_..._BASE addresses are GPU related bus addresses only. They are not valid on ARM side. The *MEM_COHERENT_REGION* range has to be mapped strongly-ordered on ARM side, so that memory accesses go straight through to the SDRAM. I used the coherent ARM alias at 0x40000000 of Alpha to realise this.

These questions are open for me now:

* How can I send a string to Alpha to be displayed on the GDB console?

* Do I need to do something so that Ctrl-C is working to stop the execution? This is not working yet.

Julio-Guerra:
Good job.

> • How can I send a string to Alpha to be displayed on the GDB console?

It should be pretty easy and straightforward. You basically just need to use the newlib as you are used to, but stub its syscalls with ours. These syscalls call Alpha to delegate them to the client by using GDB File I/O requests.

[Here is the list of syscalls supported by this GDB](https://sourceware.org/gdb/onlinedocs/gdb/List-of-Supported-Calls.html#List-of-Supported-Calls). It is enough to perform a crazy amount of things bare-metal while developping like logging, testing, mocking...

It is showcased in our [Hello World example](https://github.com/farjump/raspberry-pi/blob/master/src/hello-world/HelloWorld.c). It is compiled using the newlib from the toolchain but its syscalls are stubbed using a [GCC spec file](https://github.com/farjump/raspberry-pi/blob/master/sdk/Alpha.specs) using gcc option `-specs=sdk/Alpha.specs`. The reason why we use a spec file is to allow our users to use any `arm-none-eabi` toolchain they want (compiled with the newlib), since it modifies how the libc is linked by inserting `-lfileio -lalpha -larm`. But of course you can also explicitly link with these libraries from the command line.


> • Do I need to do something so that Ctrl-C is working to stop the execution? This is not working yet.

Alpha enables the UART interrupts to make this possible. I tried unmasking external interrupts in the CPSR register (GDB command `set $cpsr &= ~0x80` ;-)) but it didn't help. So maybe Circle reprograms it?

Let me know if you need help on these two.

rsta2:
Ctrl-C is working now. IRQs from the Mini UART will be transferred from Circle to the Alpha stub IRQ handler.

Unfortunately I was not able to write to the GDB console yet. I tried several things with "swi", but without success. The system always freezes, when the "swi" is called. I would need some info about Alpha's file IO API interface (calling method, parameters, do I have to do open() first?) for writing text to the GDB console, if this feature is wanted.

If you want to enable the Alpha GDB server support in *Config.mk* now, you have to add:

```
CFLAGS += -DUSE_ALPHA_STUB_AT=0x7F00000
AFLAGS += -DUSE_ALPHA_STUB_AT=0x7F00000
OPTIMIZE = -O0 -g3 -ggdb
```

But I would recommend to use the system option *USE_ALPHA_STUB_AT* in *include/circle/sysconfig.h* and to enable it there.

Julio-Guerra:
Regarding the UART IRQ, I am totally fine with your solution, but we usually avoid modifying the software by using GDB to prepare callbacks to Alpha from any function modifying exception/interrupt entries. This callback to Alpha restores again the exception table entries and Alpha would then chain back exceptions to the original entries when required.

If you want to go this way, use [this GDB macro](https://github.com/farjump/raspberry-pi/blob/master/sdk/alpha.gdb#L9):
```
tbreak *CExceptionHandler::CExceptionHandler
command
alpha_hook_write_restore_exception_table
continue
end

tbreak *CInterruptSystem::Initialize
command
alpha_hook_write_restore_exception_table
continue
end
```

Regarding the File I/Os, its easier than that if you simply use the [libalpha](https://github.com/farjump/raspberry-pi/blob/master/sdk/libalpha.a). I added (https://github.com/farjump/raspberry-pi/commit/2703f7b03926fff6a4a8946789da869a05cad33d) [the source code of the newlib syscalls](https://github.com/farjump/raspberry-pi/tree/master/sdk/libfileio/src) so that you can have an example of every File I/O function.

rsta2:
I can revert the last commit on the *alpha-gdbstubs-support* branch, if you prefer your solution restoring the exception vectors. It's a little bit more complicated to be used, but more compatible. For example if you decide to use the PL011 UART instead of the Mini UART, my solution would not work any more.

I wrote a string to the GDB console successfully using `write (1, "Hello GDB!\n", 11);` and *libfileio.a*. So this is working.

Unfortunately your Freeware license is not compatible with the GPLv3 (e.g. it does not allow commercial usage, GPL does), how it seems to me, and I normally avoid using external libraries in Circle itself, despite libgcc.a, which comes with the toolchain.

But a user, which wants to use your File I/O functions can easily install and use your SDK by himself. Normally you can read the Circle log messages on the screen, so this is not a problem.

rsta2:
I have tried *sample/21-webserver* which uses timer IRQs and IRQs from the USB host controller with the new *alpha.gdb* solution. Unfortunately it does not work. Ctrl-C does not stop the program and I don't know the reason. Because you have provided this solution and I'm not familiar with GDB script programming, it would be good, if you can try to solve this.

I would prefer the Alpha specific *CDevice* target for *CLogger*.

This ARM timer feature is really useful.

rsta2:
Ctrl-C is working now here with your solution. It was my own fault. I'm sorry. I was not aware of the difference of `tbreak *CInterruptSystem::Initialize` (breaks at function entry) and `tbreak CInterruptSystem::Initialize()` (breaks inside the function) and used the second one. Still learning GDB. I used the second variant, because with the automatic code completion my first try was `tbreak *CInterruptSystem::Initialize()` which was resulting in an error. So I removed the wrong token "*" instead of "()". :(

Julio-Guerra:
Really? I couldn't make it work on my side, and I am using your branch before the last commit with the UART IRQ chaining (d1c423a3), on a RPi 3. Can you show your GDB script? And regarding the `*function` notation, the only reason why we say to use it is to make sure the link register is untouched with the exact value from the call. So both notations should usually work.

Regarding the File I/O functions, I implemented the logger target but it seems to corrupt the stack, I am still investigating on it.

rsta2:
I have successfully tested it again with the commit, you have noted. This is my GDB script:

```
source alpha.gdb

# Connect to the RPi
set serial baud 115200
target remote /dev/ttyUSB0

# Load the executable in the target
# By default, the loaded file is the one GDB debugs, given as argument
# to gdb or using the command file.
load

#tbreak *CExceptionHandler::CExceptionHandler
#command
#alpha_hook_write_restore_exception_table
#continue
#end

tbreak *CInterruptSystem::Initialize
command
alpha_hook_write_restore_exception_table
continue
end

# Run untile reaching main
tbreak main
continue
```

*alpha.gdb* is the same as referenced in your comment above. I have out-commented the `tbreak *CExceptionHandler::CExceptionHandler` call, because CExceptionHandler does not modify the exception table with *USE_ALPHA_STUB_AT* defined, which also defines *USE_RPI_STUB_AT*. But it is working with this call too, so this cannot be the reason.

This is my Config.mk:

```
RASPPI = 2
CFLAGS += -DUSE_ALPHA_STUB_AT=0x7F00000
OPTIMIZE = -O0 -g3 -ggdb
```

Regarding the `*function` notation, this was exactly my problem. Because you modify the LR register in your GDB function, this function does not work, when LR is already pushed on the stack, which happens without the leading asterisk.

Julio-Guerra:
We have the same scripts, but we were not playing with the same hardware 😄So I tested on every models I have:
- the GDB-based hook to Alpha works on the RPi 1 and 2, but not on the RPi 3.
- calls to the File I/O functions of our lib only works on the RPi 1.

Good hints to solve these problems . I'll let you know as soon as we figure them out.

rsta2:
Are there any news on this? I have tested the GDB-based hook to Alpha successfully on the RPi 3B+.

rsta2:
That looks good.

Unfortunately I get the following error when doing *make* in *addon/alpha*:

```
In file included from gdbconsole.cpp:2:0:
fileio.h:13:10: error: 'fio_uint32_t' does not name a type
   static fio_uint32_t m_nErrno;
          ^~~~~~~~~~~~
```

I think, the type *fio_int32_t* is missing too. They are not defined in Circle or in the Alpha SDK. There may be another problem, because there are two files referenced by <alpha/fileio.h> (one in the SDK and one in *addon/alpha*.

Julio-Guerra:
On this one, we will have to wait until this GDB patch is merged: https://sourceware.org/ml/gdb-patches/2018-04/msg00589.html

I'm adding it in in Alpha's SDK folder until it's done.

rsta2:
Now it builds and the sample does some interesting things. But I get a panic message at some point. Here is the log:

```
logger: Circle 34.1 started on Raspberry Pi 2 Model B
kernel: Hello from CLogger
kernel: Compile time: May  2 2018 11:03:41
kernel:  Exemple 0  open(): create and open file `gdb-fileio-example.svg`
kernel:  Exemple 1  write(): write an SVG image in `gdb-fileio-example.svg`
kernel:  Exemple 2  close(): close the file descriptor of `gdb-fileio-example.svg`
kernel:  Exemple 3  rename(): rename `gdb-fileio-example.svg` into `farjump.svg`
kernel:  Exemple 4  system(): run a shell script to check the existence of file `farjump.svg`
+ echo using system
using system
+ test -e farjump.svg
kernel:  Exemple 5  stat(): get and print the file status of `farjump.svg`
File mode:                0
I-node number:            302160896
Device ID:                0x0
Container Device ID:      0x0
Mode:                     24440200000 (octal)
Link count:               16777216
Ownership:                UID=-402456576   GID=1677721600
Preferred I/O block size: 1048576 bytes
File size:                -1534918656 bytes
Blocks allocated:         1207959552
Last status change:       5259379
Last file access:         260041050
Last file modification:   260041050
kernel:  Exemple 6  open(): open file `farjump.svg` in read-only mode
kernel:  Exemple 7  read(): partly read file `farjump.svg` and compare the retrieved data with the original data
kernel:  Exemple 8  lseek(): partly read file `farjump.svg` and compare the retrieved data with the original data
file cursor moved at 159kernel:  Exemple 9  isatty(): check that `farjump.svg` is indeed not a tty
kernel:  Exemple 10  close(): closing file `farjump.svg`
kernel:  Exemple 11  isatty(): check that stdin is a TTY
kernel:  Exemple 12  unlink(): removing file `farjump.svg`
kernel:  Exemple 13  gettimeofday(): display current date and time
GMT-0 time in seconds is 260041050
kernel:  Exemple 14  using the File I/O extension with special file `/dev/random`
open(): open special file `/dev/random`
kernel: open(): �

Program received signal SIGTRAP, Trace/breakpoint trap.
CLogger::WriteV (this=0x227ed0, pSource=0x14564 "kernel", Severity=LogPanic, pMessage=0x504520 "open():", <incomplete sequence \352>, Args=...) at logger.cpp:162
162                     Breakpoint (0);
```

I think, the file stat info isn't OK.

Julio-Guerra:
`open()` on `/dev/random` failed. Do you have it on your operating-system?

Regarding the stat struct, this is an endianness problem I am solving right now: GDB's communication protocol is big endian and I am going to byte-swap structures in our library. I'll let you know when this is online.

rsta2:
Yes, I have */dev/random*. I'm using 64-bit Linux. FileIO::m_nErrno is 19 after this call to open(). I think, the format specifier ("%s") is wrong in some panic messages.

christophefarjump:
Hello, 

I am the Alpha designer, I have make new Alpha software debugger for Circle, one for RPI1 and one for RPI2 and RPI3.
Both Alpha can debug every kernel.elf or kernel7.elf present in Circle/sample directory.
And I do not need to modify anything in Circle sources,  not even need to set  USE_ALPHA_STUB_AT or USE_RPI_STUB_AT. 
Alpha debug the realese version of Circle.
To do this, i break with Alpha in strategic point in Circle startup, like MMU, exception, interrupt initialisation to make a dynamic modification of Circle system (with gdb script) to continue the debug session after Circle initalisation.
And I have a last issue and I request your expertise.
As Alpha is starting first in RPI, i must choose an address for its code and data. Alpha need 256KB of continuous RAM.
I would like to know where is the best placeto reserve this 256KB of RAM for Alpha in your Circle system ?

rsta2:
You can safely use the last megabyte of the available ARM memory for Alpha. It is not used by Circle. To determine its location you should call [this property mailbox function](https://github.com/raspberrypi/firmware/wiki/Mailbox-property-interface#get-arm-memory), subtract 1 MB from the size result and take the resulting value as your base address. The base address returned by this property mailbox function is always zero.

How is your project related to the one of @Julio-Guerra ? I was still waiting for some modifications by him to complete Alpha support in Circle. Is this obsolete now?

Julio-Guerra:
> How is your project related to the one of @Julio-Guerra ? I was still waiting for some modifications by him to complete Alpha support in Circle. Is this obsolete now?

No, it is not obsolete :) The approach Christophe talks about was more or less explained earlier in the discussion of this PR: we normally don't modify the source code to be able to debug it. So when I explained to Christophe the problems I had with Circle (FileIO and Ctrl-C problems), he wanted to get a closer look and see how we could debug Circle with our usual approach. And everything seems to be working now (there was a bug found in the interrupt handler).

So as soon as Christophe relocates Alpha somewhere else, we will update this PR with everything new, and let you know.

rsta2:
OK. Let me know.

christophefarjump:
Hi,

I have an issue with your proposal of using the property mailbox function. Alpha is executed first by the GPU loader. I must know the linking address for Alpha. Once Alpha run at an address, it can call property mailbox function but it is too late to change the its own code address.
Can i make assumptions about the GPU memory used with Circle, is it always the default value, i.d. 64 MB ?
So with RPI1 with 512MB, I can use for Alpha the address, 512MB - 64 MB - 1MB = 447MB. And with an RPI2 ou RPI3 with 1GB, i can use gor Alpha the address 1024 - 64MB - 1Mb = 959 MB ?

rsta2:
I understand the problem. But the amount of memory, which is reserved for the GPU can be configured in *config.txt* with *gpu_mem=*. So the base address for Alpha would change.

There is another option. You can use use 1 MB  at 0x300000. There is a hole in the Circle memory map at this address. This will not work, if the system option *KERNEL_MAX_SIZE* in *include/circle/sysconfig.h* is changed, but should be normally OK.

christophefarjump:
The philosophy of Alpha is to not require any modification of the debugged code.
Or at least, the minimum of change.
So at the beginning of this summer, i made a Alpha software for raspberry which have the knowlegde of circle. 
With this Alpha, you can debug the master branch of Circle without any modification 
I propose you to test it: https://github.com/farjump/Alpha_Raspberry_Pi_Circle.
I hope there will not be some regression with a new Circle Software.
Let me know if you have some issue because this Alpha had work on my environment.

rsta2:
Thanks for info about the new Alpha version. I will test it and let you know the results.

rsta2:
I have tested the Circle-specific version of Alpha on a RPi 3B+ with three different sample programs and everything worked well. Good job! It's especially useful, that the kernel image does not need to be prepared for Alpha. Each image should work.

So there is no need for a special support for debugging with Alpha in Circle any more and I think, this pull request can be closed.

rsta2:
Thanks for providing this version of Alpha! I will add an issue to your project on GitHub in case I see any incompatibilities with future versions of Circle.

PR #84, opened 2018-10-13
Applied fix from Linux touchscreen driver

stephanedamo:
Added nEventID which checks for event type.


rsta2:
I have merged your commit to the *develop* branch of Circle and did a quick successful test. It will be on master with the next release, which will probably come in October. I also did a small modification to ensure, that the touch screen driver is still working if the enum *TTouchScreenEvent* will change in the future, which is not out of question. Thanks for this contribution and for all the testing!

PR #96, opened 2019-03-03
make 'make' command externally controllable

conversy:
This allows for this kind of compile command:
env MAKE="gmake" ./makeall
(useful on mac, since system make version is < 4)

or even better:
env MAKE="gmake -j" ./makeall
for fast compile


rsta2:
This makes sense. I have cherry-picked your commit on the *develop* branch, where development takes place, and pushed it to GitHub. Thanks!

PR #99, opened 2019-07-06
Expose the device ID (SD card ID) in CEMMCDevice

stephanedamo:


stephanedamo:
Well, I have no idea why Github merged my two separate push in this pull request (for the backlight brightness) but anyway.
Backlight level seems to take a range from 0 to 180 but I haven't found the exact value. Not sure if it should be in CScreenDevice, it's releated to the screen but it doesn't use the class members it's just a mailbox call.

rsta2:
I merged this manually, because I think, the backlight setting is more low-level than CScreenDevice and fits better into CBcmFrameBuffer. There may be applications, which do not use CScreenDevice, but want to set the backlight level. Thanks for this contribution!

PR #133, opened 2020-09-13
Add note about restarting RPi to re-run bootloader

### LINKED ISSUE #132: Help on using the bootloader
Hi,

My problem is that the bootloader works only once after copying the files to the SD. The uploading doesn't work the second time I try it.

The Config.mk file contains the following lines:
```
SERIALPORT = /dev/ttyUSB1
FLASHBAUD = 115200
USERBAUD = 115200
```
My compiler machine is a PC with Ubuntu.
My Pi is 1B model.

These are the steps I follow:

- Copy all the files from circle/boot to the SD except config.txt
- Power on the Pi. A kernel which shows a RGB image is run.
- Build the 02-screenpixel sample.
- ```make flash``` from there.
- The bootloader runs normally and I see in the USB to serial adapter that the transmitting led is blinking.
- After completion the sample program is run on the Pi. The Pi green LED blinks as it should, and the rectangle with a cross is shown on the screen.
- Then I try to make a change in the sample program, build again and ```make flash``` again.
- From the PC side all is normal and there are no errors, and the UART adapter led blinks.
- But from the Pi side, it doesn't seem affected. The Pi green LED continues to blink and the cross and rectangle keep showing.

Expected result: In the last step, the Pi should reboot and execute the program with the latest changes.

I have followed all the instructions like building the libraries with ```./makeall``` and building the bootloader with ```make``` in ```/boot```.

Thanks in advance.

yomboprime:
Fixes #132

### DIFF
--- a/doc/bootloader.txt
+++ b/doc/bootloader.txt
@@ -51,3 +51,6 @@ sample program) in a shell and enter:
 terminal program "putty" directly from the shell. You have to enter "make
 monitor" like you have done for "flash" before and putty should open with the
 right communication parameters.
+
+8. To start another development cycle, power off and on the Raspberry Pi, and
+after rebuilding do again "make flash".


rsta2:
Thanks!

PR #138, opened 2020-09-23
boot: Download "cutdown" start.elf variants

dwhinham:
The Raspberry Pi firmware repository contains "cutdown" versions of
start.elf/start4.elf (start_cd.elf/start4cd.elf) that don't include
codec/3D-related code.

These versions may be useful for those looking to reduce the Raspberry
Pi boot time to an absolute minimum, as they are smaller and quicker to
load.


rsta2:
Thanks! I will add a notice to *boot/README* on how the cutdown firmware can be used.

dwhinham:
Thankyou for merging! I see the `boot/README` update too, perfect. 😄

PR #144, opened 2020-10-18
support for USB mouse HID report protocol

hinxx:
This can be considered an initial PR attempt as the code can be further improved and debugged, but I'm throwing it out there for others to tests and comment. Also the coding style might not match the rest of the code base.
Some code, specifically `TMouseReport` storage and its use in the `CUSBMouseDevice::ReportHandler()`, and the bit to integer extraction code `CUSBMouseDevice::ExtractUnsigned()` should be looked into.
Major parts and ideas were adopted from [this driver](https://github.com/rsta2/circle/blob/master/lib/usb/usbgamepadstandard.cpp) and that brings the possibility of creating a more generic HID report parsing, that could be used by all USB HID devices in the future. One thing to note is that the HID report descriptors are quite versatile and flexible as I found out in the last couple of days, and parsing them out in a safe and generic way might be a quite a challenge. So far, I tested this with three different USB mice, all with differing HID report descriptor content and layout.

A a direct consequence of the this change, support for mouse wheel is demonstrated in the updated 10-usbmouse sample. 

Here are the HID report descriptors I tested this with [parsed.log](https://github.com/rsta2/circle/files/5397622/parsed.log).

Here is the example serial dump of the detected USB mouse with the HID report descriptor and parsed report items:
```
logger: Circle 43.1 started on Raspberry Pi Zero W                                     
00:00:01.00 timer: SpeedFactor is 1.00                                                 
00:00:01.36 usbdev0-1: Device ven248a-8367 found
00:00:01.37 usbdev0-1: Interface int3-1-2 found
00:00:01.37 usbdev0-1: Using device/interface int3-1-2
00:00:01.38 usbdev0-1: Interface int3-1-1 found
00:00:01.38 usbdev0-1: Using device/interface int3-1-1
00:00:01.44 umouse: Report descriptor
00:00:01.45 umouse: Dumping 0x8E bytes starting at 0x50B120
00:00:01.45 umouse: B120: 05 01 09 02 A1 01 85 01-09 01 A1 00 05 09 19 01
00:00:01.46 umouse: B130: 29 05 15 00 25 01 95 05-75 01 81 02 95 01 75 03
00:00:01.46 umouse: B140: 81 01 05 01 09 30 09 31-15 81 25 7F 75 08 95 02
00:00:01.47 umouse: B150: 81 06 09 38 15 81 25 7F-75 08 95 01 81 06 C0 C0
00:00:01.48 umouse: B160: 05 0C 09 01 A1 01 85 03-75 10 95 02 15 01 26 8C
00:00:01.48 umouse: B170: 02 19 01 2A 8C 02 81 00-C0 05 01 09 80 A1 01 85
00:00:01.49 umouse: B180: 04 75 02 95 01 15 01 25-03 09 82 09 81 09 83 81
00:00:01.50 umouse: B190: 60 75 06 81 03 C0 05 01-09 00 A1 01 85 05 06 00
00:00:01.50 umouse: B1A0: FF 09 01 15 81 25 7F 75-08 95 07 B1 02 C0 A2 D2
00:00:01.51 umouse: Report ID 1, size 5 bytes, item count 4
00:00:01.52 umouse: [0] item ID 1 offset 8, count 5
00:00:01.52 umouse: [1] item ID 2 offset 16, count 8
00:00:01.52 umouse: [2] item ID 3 offset 24, count 8
00:00:01.53 umouse: [3] item ID 4 offset 32, count 8
00:00:01.54 dwroot: Device configured
00:00:01.54 kernel: Compile time: Oct 18 2020 15:36:14
00:00:01.55 kernel: Mouse attached
```

Looking forward to your comments and ideas!

IMG_8910

### DIFF
--- a/include/circle/input/mouse.h
+++ b/include/circle/input/mouse.h
@@ -28,7 +28,7 @@
 #define MOUSE_DISPLACEMENT_MIN	-127
 #define MOUSE_DISPLACEMENT_MAX	127
 
-typedef void TMouseStatusHandler (unsigned nButtons, int nDisplacementX, int nDisplacementY);
+typedef void TMouseStatusHandler (unsigned nButtons, int nDisplacementX, int nDisplacementY, int nWheelMove);
 
 class CMouseDevice : public CDevice	/// Generic mouse interface device ("mouse1")
 {
@@ -65,7 +65,7 @@ public:
 
 public:
 	/// \warning Do not call this from application!
-	void ReportHandler (unsigned nButtons, int nDisplacementX, int nDisplacementY);
+	void ReportHandler (unsigned nButtons, int nDisplacementX, int nDisplacementY, int nWheelMove);
 
 private:
 	CMouseBehaviour m_Behaviour;
--- a/include/circle/input/mouse.h
+++ b/include/circle/input/mouse.h
@@ -33,7 +33,10 @@ typedef void TMouseStatusHandler (unsigned nButtons, int nDisplacementX, int nDi
 class CMouseDevice : public CDevice	/// Generic mouse interface device ("mouse1")
 {
 public:
-	CMouseDevice (void);
+	/// \brief Construct a mouse device instance
+	/// \param nButtons Number of buttons
+	/// \param bHasWheel Wheel presence
+	CMouseDevice (unsigned nButtons, boolean bHasWheel = FALSE);
 	~CMouseDevice (void);
 
 	/// \brief Setup mouse device in cooked mode
@@ -63,6 +66,12 @@ public:
 	/// \param pStatusHandler Pointer to the mouse status handler
 	void RegisterStatusHandler (TMouseStatusHandler *pStatusHandler);
 
+	/// \return Number of supported mouse buttons
+	unsigned GetButtonCount (void) const;
+
+	/// \return Does the mouse support a mouse wheel?
+	boolean HasWheel (void) const;
+
 public:
 	/// \warning Do not call this from application!
 	void ReportHandler (unsigned nButtons, int nDisplacementX, int nDisplacementY, int nWheelMove);
@@ -74,6 +83,9 @@ private:
 
 	unsigned m_nDeviceNumber;
 	static CNumberPool s_DeviceNumberPool;
+
+	unsigned m_nButtons;
+	boolean m_bHasWheel;
 };
 
 #endif
--- a/include/circle/input/mousebehaviour.h
+++ b/include/circle/input/mousebehaviour.h
@@ -29,7 +29,7 @@ enum TMouseEvent
 	MouseEventMouseUp,
 	//MouseEventClick,
 	//MouseEventDoubleClick,
-    MouseEventMouseWheel,
+	MouseEventMouseWheel,
 	MouseEventUnknown
 };
 
--- a/include/circle/input/mousebehaviour.h
+++ b/include/circle/input/mousebehaviour.h
@@ -29,6 +29,7 @@ enum TMouseEvent
 	MouseEventMouseUp,
 	//MouseEventClick,
 	//MouseEventDoubleClick,
+    MouseEventMouseWheel,
 	MouseEventUnknown
 };
 
@@ -36,7 +37,7 @@ enum TMouseEvent
 #define MOUSE_BUTTON_RIGHT	(1 << 1)
 #define MOUSE_BUTTON_MIDDLE	(1 << 2)
 
-typedef void TMouseEventHandler (TMouseEvent Event, unsigned nButtons, unsigned nPosX, unsigned nPosY);
+typedef void TMouseEventHandler (TMouseEvent Event, unsigned nButtons, unsigned nPosX, unsigned nPosY, int nWheelMove);
 
 class CMouseBehaviour
 {
@@ -54,7 +55,7 @@ public:
 	void UpdateCursor (void);	// call this frequently from TASK_LEVEL
 
 public:
-	void MouseStatusChanged (unsigned nButtons, int nDisplacementX, int nDisplacementY);
+	void MouseStatusChanged (unsigned nButtons, int nDisplacementX, int nDisplacementY, int nWheelMove);
 
 private:
 	static boolean SetCursorState (unsigned nPosX, unsigned nPosY, boolean bVisible);
--- a/include/circle/usb/usbmouse.h
+++ b/include/circle/usb/usbmouse.h
@@ -24,6 +24,30 @@
 #include <circle/input/mouse.h>
 #include <circle/types.h>
 
+enum TMouseReportType {
+    MouseItemNone,
+    MouseItemButtons,
+    MouseItemXAxis,
+    MouseItemYAxis,
+    MouseItemWheel
+};
+
+struct TMouseReportItem {
+    unsigned type;
+    unsigned count;
+    unsigned offset;
+};
+#define MAX_ITEMS    16
+
+struct TMouseReport
+{
+    unsigned id;
+    unsigned size;
+
+    unsigned nItems;
+    TMouseReportItem items[MAX_ITEMS];
+};
+
 class CUSBMouseDevice : public CUSBHIDDevice
 {
 public:
@@ -34,12 +58,17 @@ public:
 
 private:
 	void ReportHandler (const u8 *pReport, unsigned nReportSize);
+    void DecodeReport (void);
+    u32 ExtractUnsigned (const void *buffer, u32 offset, u32 length);
+	s32 ExtractSigned (const void *buffer, u32 offset, u32 length);
 
 private:
 	CMouseDevice *m_pMouseDevice;
 
 	u8 *m_pHIDReportDescriptor;
 	u16 m_usReportDescriptorLength;
+
+    TMouseReport m_ReportItems;
 };
 
 #endif
--- a/include/circle/usb/usbmouse.h
+++ b/include/circle/usb/usbmouse.h
@@ -24,28 +24,27 @@
 #include <circle/input/mouse.h>
 #include <circle/types.h>
 
-enum TMouseReportType {
-	MouseItemNone,
+enum TMouseReportType
+{
 	MouseItemButtons,
 	MouseItemXAxis,
 	MouseItemYAxis,
-	MouseItemWheel
+	MouseItemWheel,
+
+	MouseItemCount
 };
 
-struct TMouseReportItem {
-	unsigned type;
-	unsigned count;
-	unsigned offset;
+struct TMouseReportItem
+{
+	unsigned bitSize;
+	unsigned bitOffset;
 };
-#define MAX_MOUSE_REPORT_ITEMS    16
 
 struct TMouseReport
 {
 	unsigned id;
-	unsigned size;
-
-	unsigned nItems;
-	TMouseReportItem items[MAX_MOUSE_REPORT_ITEMS];
+	unsigned byteSize;
+	TMouseReportItem items[MouseItemCount];
 };
 
 class CUSBMouseDevice : public CUSBHIDDevice
@@ -58,7 +57,7 @@ public:
 
 private:
 	void ReportHandler (const u8 *pReport, unsigned nReportSize);
-	void DecodeReport (unsigned *nButtons, boolean *bHasWheel);
+	void DecodeReport (void);
 	u32 ExtractUnsigned (const void *buffer, u32 offset, u32 length);
 	s32 ExtractSigned (const void *buffer, u32 offset, u32 length);
 
--- a/include/circle/usb/usbmouse.h
+++ b/include/circle/usb/usbmouse.h
@@ -58,7 +58,7 @@ public:
 
 private:
 	void ReportHandler (const u8 *pReport, unsigned nReportSize);
-	void DecodeReport (void);
+	void DecodeReport (unsigned *nButtons, boolean *bHasWheel);
 	u32 ExtractUnsigned (const void *buffer, u32 offset, u32 length);
 	s32 ExtractSigned (const void *buffer, u32 offset, u32 length);
 
--- a/include/circle/usb/usbmouse.h
+++ b/include/circle/usb/usbmouse.h
@@ -37,7 +37,7 @@ struct TMouseReportItem {
 	unsigned count;
 	unsigned offset;
 };
-#define MAX_ITEMS    16
+#define MAX_MOUSE_REPORT_ITEMS    16
 
 struct TMouseReport
 {
@@ -45,7 +45,7 @@ struct TMouseReport
 	unsigned size;
 
 	unsigned nItems;
-	TMouseReportItem items[MAX_ITEMS];
+	TMouseReportItem items[MAX_MOUSE_REPORT_ITEMS];
 };
 
 class CUSBMouseDevice : public CUSBHIDDevice
@@ -68,7 +68,7 @@ private:
 	u8 *m_pHIDReportDescriptor;
 	u16 m_usReportDescriptorLength;
 
-	TMouseReport m_ReportItems;
+	TMouseReport m_MouseReport;
 };
 
 #endif
--- a/include/circle/usb/usbmouse.h
+++ b/include/circle/usb/usbmouse.h
@@ -25,27 +25,27 @@
 #include <circle/types.h>
 
 enum TMouseReportType {
-    MouseItemNone,
-    MouseItemButtons,
-    MouseItemXAxis,
-    MouseItemYAxis,
-    MouseItemWheel
+	MouseItemNone,
+	MouseItemButtons,
+	MouseItemXAxis,
+	MouseItemYAxis,
+	MouseItemWheel
 };
 
 struct TMouseReportItem {
-    unsigned type;
-    unsigned count;
-    unsigned offset;
+	unsigned type;
+	unsigned count;
+	unsigned offset;
 };
 #define MAX_ITEMS    16
 
 struct TMouseReport
 {
-    unsigned id;
-    unsigned size;
+	unsigned id;
+	unsigned size;
 
-    unsigned nItems;
-    TMouseReportItem items[MAX_ITEMS];
+	unsigned nItems;
+	TMouseReportItem items[MAX_ITEMS];
 };
 
 class CUSBMouseDevice : public CUSBHIDDevice
@@ -58,8 +58,8 @@ public:
 
 private:
 	void ReportHandler (const u8 *pReport, unsigned nReportSize);
-    void DecodeReport (void);
-    u32 ExtractUnsigned (const void *buffer, u32 offset, u32 length);
+	void DecodeReport (void);
+	u32 ExtractUnsigned (const void *buffer, u32 offset, u32 length);
 	s32 ExtractSigned (const void *buffer, u32 offset, u32 length);
 
 private:
@@ -68,7 +68,7 @@ private:
 	u8 *m_pHIDReportDescriptor;
 	u16 m_usReportDescriptorLength;
 
-    TMouseReport m_ReportItems;
+	TMouseReport m_ReportItems;
 };
 
 #endif
--- a/lib/input/mouse.cpp
+++ b/lib/input/mouse.cpp
@@ -77,12 +77,12 @@ void CMouseDevice::RegisterStatusHandler (TMouseStatusHandler *pStatusHandler)
 	assert (m_pStatusHandler != 0);
 }
 
-void CMouseDevice::ReportHandler (unsigned nButtons, int nDisplacementX, int nDisplacementY)
+void CMouseDevice::ReportHandler (unsigned nButtons, int nDisplacementX, int nDisplacementY, int nWheelMove)
 {
-	m_Behaviour.MouseStatusChanged (nButtons, nDisplacementX, nDisplacementY);
+	m_Behaviour.MouseStatusChanged (nButtons, nDisplacementX, nDisplacementY, nWheelMove);
 
 	if (m_pStatusHandler != 0)
 	{
-		(*m_pStatusHandler) (nButtons, nDisplacementX, nDisplacementY);
+		(*m_pStatusHandler) (nButtons, nDisplacementX, nDisplacementY, nWheelMove);
 	}
 }
--- a/lib/input/mouse.cpp
+++ b/lib/input/mouse.cpp
@@ -26,9 +26,11 @@ CNumberPool CMouseDevice::s_DeviceNumberPool (1);
 static const char FromMouse[] = "mouse";
 static const char DevicePrefix[] = "mouse";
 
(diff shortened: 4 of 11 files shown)


rsta2:
That was quick. I tested another two mice and both are working with your modified driver and sample/10-usbmouse. Great! I need some time for evaluation.

rsta2:
I had a look at your changes and here are my comments. First thank you for finding out, how the mouse wheel works and implementing these changes. To make this part of Circle I want to request some more modifications to your code:

* You cannot set REPORT_PROTOCOL unconditionally for 3-1-X devices in usbhiddevice.cpp. This will not work with USB keyboards then. There can be another parameter to the constructor of CUSBHIDDevice. Only if this parameter is TRUE, REPORT_PROTOCOL will be set:

```
class CUSBHIDDevice
{
...
	CUSBHIDDevice (CUSBFunction *pFunction, unsigned nMaxReportSize = 0,
			boolean bForceReportProtocol = FALSE);
```

* I have seen, that with report protocol there are different ranges for the mouse displacement. Previously this was fixed at -/+ 127. I think, this has to be converted in a way, if the maximum displacement is greater than this. On the API the maximum displacement should be -/+ 127 in any case.

* I have seen, that with report protocol there may be more than 3 mouse buttons. We should add these methods in <circle/input/mouse.h>:

```
class CMouseDevice
{
...
	/// \return Number of supported mouse buttons
	unsigned GetButtonCount (void) const;

	/// \return Does the mouse support a mouse wheel?
	boolean HasWheel (void) const;
```

* The define MAX_ITEMS in <circle/usb/usbmouse.h> may cause name clashes. Better would be MAX_MOUSE_REPORT_ITEMS. Normally I prefer to hide everything from the header file, which is not really needed for the interface. I know, this is not possible in all cases, but if not, the used names should be that specific, that name clashes are unlikely.

* The sample/10-usbmouse must work with any setting for the screen DEPTH (8, 16 or 32). See <circle/screen.h> for this define. Maybe the sample should only use the colors NORMAL_COLOR, HIGH_COLOR, HALF_COLOR. This should be enough to show the function. Or we define a different color handling for different DEPTHs?

* Please use at least these requirements for the coding style: Use tabs, tabs are 8 characters wide, maximum line length is 100, "{" should be on an extra line.

* The mouse descriptor and item information should not be logged in the final version. I try to be quiet here.

* The code have to be cleaned up from out-commented stuff (e.g. debug output only needed for development).

I don't know, if you want to do these changes and provide the result as a contribution for Circle? I would appreciate this.

Finally, taking into account, what you wrote about the HID report parser, I would say for the moment it is better to use a mouse specific parser inside the mouse driver. A generic parser would be great, but this is probably difficult to achieve.

hinxx:
Thanks for your comments @rsta2 !

> I don't know, if you want to do these changes and provide the result as a contribution for Circle? I would appreciate this.

Yeah, sure. Let me give it a try and update the PR in the coming days.

>  I would say for the moment it is better to use a mouse specific parser inside the mouse driver. A generic parser would be great, but this is probably difficult to achieve.

Completely agree.

> This will not work with USB keyboards then.

I just used an USB keyboard yesterday and did not notice any breakage. It was sharing the the same wireless dongle as the mouse; i.e. keyboard was not connected through it own wireless or wired interface. Maybe that makes a difference and I'll look into it.

Note to self, make sure vc4 addon samples using mouse support work too!

rsta2:
@hinxx OK, thanks!

> I just used an USB keyboard yesterday and did not notice any breakage. It was sharing the the same wireless dongle as the mouse; i.e. keyboard was not connected through it own wireless or wired interface. Maybe that makes a difference and I'll look into it.

The point is, that the USB keyboard driver is working with setting the boot protocol for long now. Because the driver has not changed and we cannot test all keyboards out there, we should remain with the boot protocol setting for the keyboard.

> Note to self, make sure vc4 addon samples using mouse support work too!

There are more samples, which use the mouse (e.g. in addon/ugui/). I will care about that. Please change the sample/10-usbmouse as described. I will do the other samples then. Thanks!

hinxx:
@rsta2 , I think I've addressed the above comments (should still trim couple of > 100 char lines, though). I'm not going to describe the changes here, code should speak for itself, but I can argument some of the design choice I made, should you need.

> You cannot set REPORT_PROTOCOL unconditionally for 3-1-X devices in usbhiddevice.cpp

I could not make this work for the mouse/keyboard setup I have. I did as you suggested and the result was unresponsive mouse. Adding a debug printout, in the serial log I could see that for one of the interfaces, `int3-1-1` and `int3-1-2`, `CUSBHIDDevice` constructor passes in `TRUE` an `FALSE` for the other. The mouse one should have been `TRUE`, as I left keyboard code intact.
As this breaks this PR I'm not sure how should we proceed. I can create another PR, based off this one, just to look into this particular situation.

rsta2:
@hinxx First thank you for applying the modifications! I haven't a look on it in detail, but I will do so soon. I did a quick test and it's working with my mouse.

Isn't is possible to set report protocol only for the mouse interface 3-1-2 and to set boot protocol for the keyboard interface 3-1-1? The control message is directed to the interface, so normally this should be supported, even with compound keyboard/mouse devices.

I tested 3 (non-compound) USB keyboards with the report protocol setting and all did work, but nevertheless this need not be the case with all keyboards. If it is really not possible to set specific protocols, that would be a problem. If we generally set report protocol and after releasing this Circle version, users noting a failure of function for their own keyboard, we cannot easily take back the change, because the API has changed by the additional wheel parameter. This may cause confusion.

So I think the only chance, if setting specific protocols does not work, would be to fully implement the report protocol for USB keyboards too, including the HID report parser. Currently I do not know, if this is possible with acceptable effort. Then the generic HID parser would come into play again.

rsta2:
I had a closer look at your modifications and that's good. Well done!

I was able to test a compound wireless keyboard/mouse from Microsoft now. Unfortunately I got an assertion failed message [here](https://github.com/hinxx/circle/blob/mouse-hid-report/lib/usb/usbmouse.cpp#L277). I removed the assert, then it's working, but the sample/10-usbmouse draws a dotted line (blue/red) then, when I press the right mouse button.

I did the following modification in usbhiddevice.cpp:

```
if (   GetInterfaceClass ()    == 3	// HID class
    && GetInterfaceSubClass () == 1)	// boot class
{
	if (GetHost ()->ControlMessage (GetEndpoint0 (),
					REQUEST_OUT | REQUEST_CLASS | REQUEST_TO_INTERFACE,
					SET_PROTOCOL,   GetInterfaceProtocol () == 2
						      ? REPORT_PROTOCOL : BOOT_PROTOCOL,
					GetInterfaceNumber (), 0, 0) < 0)
	{
		CLogger::Get ()->Write (FromUSBHID, LogError, "Cannot set boot protocol");

		return FALSE;
	}
}
```

This is working too, with boot protocol set for the keyboard, which would be good to do.

hinxx:
Glad to be able t contribute

> Isn't is possible to set report protocol only for the mouse interface 3-1-2 and to set boot protocol for the keyboard interface 3-1-1?

It should be possible. It is probably my lack of understanding the whole code stack you have in place not to see this as an option myself.

I'm with you on the whole approach on how making sure that nothing is broken with the change to the existing users is a very important item.

IMHO, and with limited USB HID knowledge, I suppose that boot protocol, being a simplification of the report protocol, is something that mice and keyboards would support to a large extent. It is used by the BIOS, AFAICT, to allow early interaction with the PC before the OS USB stack kicks in with its full fledged HID support. Also, I would suspect that the keyboard report handling might practically be a non-issue; what I've seen so far in descriptors is just number of keys that the keyboard would support, and that seems to be something that could be ignored (given that the boot mode works); except if we want to support custom/fancy keys like some keyboards might have - then the HID report might contain some info. That would explain why there is not obvious breakage noticed on my side. I might look into it a bit more in the future, but I see no need for us go there now.

Nevertheless, with all that being said, let us *not* go and assume anything about the boot/report protocols of a keyboard in the code, for the sake of not breaking things to the existing users.

> Unfortunately I got an assertion failed message here.

That assert I left in there intentionally and I'm glad it got tripped now. It would suggest that the mouse definition was found more than once in the report. It might also be the reason that by removing it, the parsing/decoding works, but you see blue/red dotted lines in the sample/10-usbmouse. The change of pixel in that sample should happen if one scrolls the wheel (no matter which direction). What your description suggests to me is, that the report was not properly parsed, and the wheel item in the stored struct is actually no a wheel (could it be that it is a position displacement instead?).
Can you, please, share the hexdump of this offending report? I would be interested to learn how Microsoft wants their HIDs to behave ;).

> This is working too, with boot protocol set for the keyboard, which would be good to do.

Agree, looks good to me too!

rsta2:
I completely agree with this things you wrote. One thing to mention: Assertions in the code are good, even in the final version. They can help very much to find bugs and they can be easily removed by defining NDEBUG. Maybe I was not clear in that before.

Here is the HID report descriptor of the USB wireless keyboard/mouse combo from Microsoft (0xDF bytes):

```
05 01 09 02 A1 01 05 01-09 02 A1 02 85 1A 09 01
A1 00 05 09 19 01 29 05-95 05 75 01 15 00 25 01
81 02 75 03 95 01 81 01-05 01 09 30 09 31 95 02
75 10 16 01 80 26 FF 7F-81 06 A1 02 85 12 09 48
95 01 75 02 15 00 25 01-35 01 45 0C B1 02 85 1A
09 38 35 00 45 00 95 01-75 10 16 01 80 26 FF 7F
81 06 C0 A1 02 85 12 09-48 75 02 15 00 25 01 35
01 45 0C B1 02 35 00 45-00 75 04 B1 01 85 1A 05
0C 95 01 75 10 16 01 80-26 FF 7F 0A 38 02 81 06
C0 C0 C0 C0 05 0C 09 01-A1 01 05 01 09 02 A1 02
85 1F 05 0C 0A 38 02 95-01 75 10 16 01 80 26 FF
7F 81 06 85 17 06 00 FF-0A 06 FF 0A 0F FF 15 00
25 01 35 01 45 0C 95 02-75 02 B1 02 0A 04 FF 35
00 45 00 95 01 75 01 B1-02 75 03 B1 01 C0 C0
```

hinxx:
Thanks for the report descriptor dump. Decoded version is [here](https://github.com/rsta2/circle/files/5408611/microsoft.log).


There's good news and not so good news.

My assumptions were pretty much spot on the reason why you are getting the red/blue pixels. Due to different (more complex) and feature packed report of this device, the parsing in this PR incorrectly handles the wheel item. In effect this causes `ReportHandler` to interpret received report packets X position as wheel movement. FWIW, that is the good news.

The not so good news is that this report contains a resolution multiplier information for X, Y position and wheel movement. Up to now I had not seen such report while working in this PR. This requires more elaborate parsing of the report to be implemented. Also, there might be a need to set the desired resolution multiplier through an additional USB request towards the device. Not sure if that is required from operational point of view (might not be an issue at all). I'm guessing that if not specifically missed until now by the users it might be avoided. We would still need to act on the currently used multiplier and apply it to position and wheel movement before reporting the numbers to the application.

Looking at this report it kind of feels like there is a stronger need to have a proper HID report parser in place in order to assure robustness with all kinds of mice out there. I would feel a bit insecure to rely on a partial implementation, knowing this might render certain mice unusable.

OTOH, the wheel feature support could be explicitly asked for from the application side, while defaulting to boot protocol.

Nevertheless I'll look into this a bit more to see how exiting PR code might handle this.
Question: If I find an existing 3rd party C/C++ code to parse the HID report, would you accept it ? What license should it be under, or better yet, under which not?

rsta2:
Thanks for the info.

There is no USB HID specific request defined in the [specification](https://www.usb.org/sites/default/files/hid1_11.pdf) (see chapter 7.2) for setting a resolution multiplier. Maybe this multiplier has to be applied on received HID report movement values only?

Requesting the wheel feature by application request is probably difficult. This is caused by the architecture of the USB stack. The USB function drivers (e.g. for the mouse) are initialized at a time, when the application does not really has access to the driver to set such information. The boot or report protocol is already set, when the application gets control. While re-setting the protocol on the device later may be possible, but the whole driver initialization would have to be changed. If possible, I would not like to do this.

Third-party code could be accepted (e.g. for implementing a class CUSBHIDReportParser). I would prefer GPL-licensed code. General info about license-compatibility is difficult to give, it depends.

hinxx:
I see how the wheel feature by application request would be difficult to provide. Just an idea I had..

> There is no USB HID specific request defined in the specification (see chapter 7.2) for setting a resolution multiplier.

I was looking at the [usage tables](https://www.usb.org/sites/default/files/hut1_12.pdf ),  chapter _4.3.1 Resolution Multiplier_. AFAICT, it would applied be through the _7.2.2 Set_Report Request_ as found in the [usb specs](https://www.usb.org/sites/default/files/hid1_11.pdf), with the Feature report type (0x03).

As the device would send the Feature report with the resolution values (report ID 18 in your case) which is different from the regular Input data (report ID 26 in your case), a need to handle such Feature reports might arise.

Thinking out loud here now. We might get away with completely ignoring the resolution business, and just make sure we parse the mouse Input report ID, and its items properly. As the mouse would report the displacement of the position the resulting effect on the application would be potentially faster/slower mouse movement depending on the value we receive from the mouse. Something I want to explore as well..

Speculation ahead: I think I just realized what the extra, small button, near the wheel, on two mice I have, does under the hood; if I click it, the mouse cursor speed increases in 3-4 stages for each click, then goes to minimum speed. What this button in effect does is change the resolution of the mouse reported data in Input report. Higher resolution = more displacement is seen by the OS. Faster (more pixels traveled) cursor movement would certainly help in getting from one side of the high resolution screen to another.. go figure.

rsta2:
How it seems, this requires some more research? Just take the time you need. Let me know, when you have questions.

hinxx:
The latest change should make your Microsoft mouse work. In this attempt the Feature items are ignored, as is the handling of the resolution multiplier altogether, while the rest of the valuable mouse items are better handled.

Let me know, please, if it works for you.

One more thing I would like to tackle before declaring readiness for inclusion from my side; handling input reports without an ID (byte 0) in the data expected from old/simple mice.

hinxx:
Going back to boot vs report protocol setting.
I still can not get one mouse to work with the change you proposed, while three other mice work just fine.
Essentially it seems that there is a strange order of detection happening in that one case. I've dumped the detection / init parts on serial port and this is what I get by using your code in `usbhiddevice.cpp`.

This mouse is OK (uses report protocol as requested):
```
00:00:01.86 usbdev0-1: Device ven62a-9013 found
00:00:01.87 usbdev0-1: Interface int3-1-1 found
00:00:01.87 usbdev0-1: Using device/interface int3-1-1
00:00:01.88 usbdev0-1: Interface int3-1-2 found
00:00:01.88 usbdev0-1: Using device/interface int3-1-2
00:00:01.94 usbhid: Using HID protocol 0, IntfNum 0, IntfProto 1, IntfClass 3, IntfSubClass 1
00:00:01.95 umouse: Report descriptor
00:00:01.96 umouse: Dumping 0xA8 bytes starting at 0x510640
00:00:01.96 umouse: 0640: 05 01 09 02 A1 01 85 03-09 01 A1 00 05 09 19 01
00:00:01.97 umouse: 0650: 29 05 15 00 25 01 75 01-95 05 81 02 75 03 95 01
00:00:01.97 umouse: 0660: 81 01 05 01 09 30 09 31-16 01 F8 26 FF 07 75 0C
00:00:01.98 umouse: 0670: 95 02 81 06 09 38 15 81-25 7F 75 08 95 01 81 06
00:00:01.99 umouse: 0680: C0 05 0C 15 81 25 7F 75-08 95 01 0A 38 02 81 06
00:00:01.99 umouse: 0690: C0 05 0C 09 01 A1 01 85-01 19 00 2A 80 03 15 00
00:00:02.00 umouse: 06A0: 26 80 03 75 10 95 01 81-00 C0 05 01 09 80 A1 01
00:00:02.01 umouse: 06B0: 85 02 19 81 29 83 25 01-75 01 95 03 81 02 95 05
00:00:02.01 umouse: 06C0: 81 01 C0 06 F2 F1 0A F2-F1 A1 01 85 05 09 00 95
00:00:02.02 umouse: 06D0: 06 75 08 15 00 26 FF 00-91 02 09 00 95 05 75 08
00:00:02.03 umouse: 06E0: 15 00 26 FF 00 81 02 C0-55 55 55 55 55 55 55 55
00:00:02.03 umouse: Report ID 3 size is 7 bytes
00:00:02.04 umouse: item ID 0 : bitoffset 8, bitsize 5
00:00:02.04 umouse: item ID 1 : bitoffset 16, bitsize 12
00:00:02.05 umouse: item ID 2 : bitoffset 28, bitsize 12
00:00:02.05 umouse: item ID 3 : bitoffset 40, bitsize 8
00:00:02.06 usbhid: Using HID protocol 1, IntfNum 1, IntfProto 2, IntfClass 3, IntfSubClass 1
00:00:02.07 dwroot: Device configured
```

This mouse is NOT OK (uses boot protocol, instead of report):
```
00:00:42.64 usbdev0-1: Device ven62a-9013 removed
00:00:42.64 usbdev0-1: Device ven248a-8367 found
00:00:42.64 usbdev0-1: Interface int3-1-2 found
00:00:42.64 usbdev0-1: Using device/interface int3-1-2
00:00:42.64 usbdev0-1: Interface int3-1-1 found
00:00:42.64 usbdev0-1: Using device/interface int3-1-1
00:00:42.64 umouse: Report descriptor
00:00:42.64 umouse: Dumping 0x8E bytes starting at 0x51A5A0
00:00:42.64 umouse: A5A0: 05 01 09 02 A1 01 85 01-09 01 A1 00 05 09 19 01
00:00:42.64 umouse: A5B0: 29 05 15 00 25 01 95 05-75 01 81 02 95 01 75 03
00:00:42.64 umouse: A5C0: 81 01 05 01 09 30 09 31-15 81 25 7F 75 08 95 02
00:00:42.64 umouse: A5D0: 81 06 09 38 15 81 25 7F-75 08 95 01 81 06 C0 C0
00:00:42.64 umouse: A5E0: 05 0C 09 01 A1 01 85 03-75 10 95 02 15 01 26 8C
00:00:42.64 umouse: A5F0: 02 19 01 2A 8C 02 81 00-C0 05 01 09 80 A1 01 85
00:00:42.64 umouse: A600: 04 75 02 95 01 15 01 25-03 09 82 09 81 09 83 81
00:00:42.64 umouse: A610: 60 75 06 81 03 C0 05 01-09 00 A1 01 85 05 06 00
00:00:42.64 umouse: A620: FF 09 01 15 81 25 7F 75-08 95 07 B1 02 C0 A2 D2
00:00:42.64 umouse: Report ID 1 size is 5 bytes
00:00:42.64 umouse: item ID 0 : bitoffset 8, bitsize 5
00:00:42.64 umouse: item ID 1 : bitoffset 16, bitsize 8
00:00:42.64 umouse: item ID 2 : bitoffset 24, bitsize 8
00:00:42.64 umouse: item ID 3 : bitoffset 32, bitsize 8
00:00:42.64 usbhid: Using HID protocol 1, IntfNum 0, IntfProto 2, IntfClass 3, IntfSubClass 1
00:00:42.64 usbhid: Using HID protocol 0, IntfNum 1, IntfProto 1, IntfClass 3, IntfSubClass 1
00:00:42.64 dwroot: Device configured

```

And here is the code I used so you can tell where the `Using HID protocol ..` lines come from:
```

	if (   GetInterfaceClass ()    == 3	// HID class
	    && GetInterfaceSubClass () == 1)	// boot class
	{
        u16 proto = 11;
        if (GetInterfaceProtocol () == 2)
            proto = REPORT_PROTOCOL;
        else if (GetInterfaceProtocol () == 1) {
            proto = BOOT_PROTOCOL;
        }
        CLogger::Get ()->Write (FromUSBHID, LogDebug, "Using HID protocol %d, IntfNum %d, IntfProto %d, IntfClass %d, IntfSubClass %d",
                                proto, GetInterfaceNumber(), GetInterfaceProtocol(), GetInterfaceClass(), GetInterfaceSubClass());
		if (GetHost ()->ControlMessage (GetEndpoint0 (),
						REQUEST_OUT | REQUEST_CLASS | REQUEST_TO_INTERFACE,
                        SET_PROTOCOL, proto,
						GetInterfaceNumber (), 0, 0) < 0)
		{
			CLogger::Get ()->Write (FromUSBHID, LogError, "Cannot set boot protocol");

			return FALSE;
		}
	}

```

rsta2:
Thanks for the latest updates. I tested my 3 mice and 3 keyboards (including Microsoft combo) with your latest version, which sets the report protocol in any case, and everything worked fine.

As you wrote, one of your mice does not work. As I understand it, this mouse is part of a keyboard/mouse combo and the keyboard interface is initialized at last to boot protocol. Maybe this means, that the whole device is working in boot protocol mode and thus the mouse, which expects to be used in report protocol mode, is not working.

I dumped the HID report descriptor of my 3 keyboards and it's true, that these descriptors do not tell much.

Are you secure with the HID report descriptor parser for the mouse now?

hinxx:
>  I tested my 3 mice and 3 keyboards (including Microsoft combo) with your latest version, which sets the report protocol in any case, and everything worked fine.

Nice to hear!

> As you wrote, one of your mice does not work. As I understand it, this mouse is part of a keyboard/mouse combo and the keyboard interface is initialized at last to boot protocol. Maybe this means, that the whole device is working in boot protocol mode and thus the mouse, which expects to be used in report protocol mode, is not working.

I think you are spot on on all accounts. I could see mouse `ReportHandler` get 3 bytes which looks suspiciously like mouse is in boot mode.
Looking the Configuration descriptor it is clear why the order of interface initialization is the way it is for this device; the mouse vs. keyboard Interface descriptors are swapped compared to all other devices. This should not be a problem on the USB level and it is certainly nothing that Circle should worry about or handle differently, if you ask me. It is likely a device firmware issue with this mouse/keyboard combo I have and there is nothing much we an do.

> Are you secure with the HID report descriptor parser for the mouse now?

To the extent that was possible to test it, yes.

What this (short) journey into the USB HID taught me was that there is a wild, wild world out there when it comes to HID devices. And that makes me not so 100% that this parsing is robust enough. 

I found [libhid](https://openports.se/devel/libhid) which seems to lack maintainer. I would like to run couple of tests with its HID parser (it is actually the only parser I could fish out from the web) and see what happens. If it turns out more complete and robust it could also be used for the more general HID parsing, as opposed to having one off custom mouse parser. The libhid stuff would still only deal with the mouse at first, to minimize possible problems, but at least we would have a good candidate for generic parser as time goes by.

We could put this PR on hold for a while and I get back to you with the libhid findings. OTOH, if you feel that this PR adds value as-is, go ahead and merge it.

rsta2:
The commits from this PR have been merged to the *develop* branch in the Circle repository. Thus they will be part of the next release. Additionally I adapted the remaining libraries and samples for the new event handler format (wheel move added) and added more colors to *sample/10-usbmouse*. The VC4 sample *hello_triangle2* uses the mouse wheel to modify the fractal image. Thank you very much for this contribution!

PR #151, opened 2020-10-28
support for 5 mouse buttons

hinxx:
Support for a mouse that has 5 buttons; two extra buttons are on the side of the mouse operated with a thumb.
The change is pretty trivial.

As an improvement, I would like to see `CMouseBehaviour::MouseStatusChanged()` looking at the `CMouseDevice::GetButtonCount()` instead of using `MOUSE_BUTTONS` define. Can't really see how that would come about at the moment without some API changes.


rsta2:
Thanks for the PR. I think, we can let it this way for the moment. The gain is not enough for an API change.

Issue #154, opened 2020-11-04
Support for USB-serial dongles


hinxx:
I've been playing with Ryanteck RTk.GPIO  board and managed to get the basic I/O working. It looks quite useful piece of equipment if one is looking for more I/O (28 pins). I had to introduce rudimentary usbserial.cpp device driver that would talk to CH340 IC.

As a more general support of USB-serial under Circle, I'm also looking at the USB-serial converters that use different popular ICs (CP2102, FT245R, PL2303). As bare minimum support for baud rate, line control (# data bits, # stop bits, parity) and possibly HW flow control would be implemented.

Does it look interesting enough to introduce usbserial to Circle?
In your opinion, how would be best to handle such differences at IC level when introducing USB-serial support? As different IC devices drivers with overarching usbserial layer, or maybe leave the specifics to the app?

rsta2:
Yes, USB serial support would be great to have in Circle. Unfortunately I'm not familiar with the differences of the different chips.

Normally I would suggest, to define a unique base class (e.g. *CUSBSerialDevice*), which is derived from *CUSBFunction* and defines the API to the serial interfaces, which should be basically the *Read/Write()* interface of *CDevice* (*CUSBFunction* is already derived from *CDevice*) with some extensions for parameter setting. Currently not sure, how flow control can be realized. The specific chip driver classes can be derived from *CUSBSerialDevice* then and implemented separately. I would try to hide most chip specifics in these specific driver classes, but I don't know, if this is possible with reasonable effort.

There will probably a problem arise with the handling of many different vendor/device ID combinations. I have seen huge lists of this in some Linux driver. I suppose, the current matching method in *usbdevicefactory.cpp* is not suitable for this. We will need an other solution, may be as an extension of the current one.

hinxx:
Perfect!

So far, I did the exact same thing that you mention in 2nd paragraph so far in introducing _CUSBSerialDevice_ :). Your suggestion, where specific IC device support would derive _CUSBSerialDevice_, and implement its interfaces, sounds very good to me. I have couple of dongles based on different ICs and will try to figure out how they would support the minimal set of functionality.

[...]

PR #155, opened 2020-11-04
Usb to serial support in Circle

hinxx:
Here is an initial attempt at adding USB to serial support for a RTK GPIO board. The board is based on Winchiphead CH340 IC. The low-level chip register addresses, baud rate and framing configuration was taken from Linux kernel `ch341.c` device driver.
Sample program is also provided. The code was tested on an actual RTK GPIO board (https://github.com/PiSupply/Ryanteck/tree/master/RTK.GPIO).

The approach taken with the _CUSBSerialDevice_ and _CUSBSerialCH341Device_ allows adding support for other USB to serial devices, based on other ICs, as discussed in #154 . Support for those will be coming soon.

Comments welcomed.


rsta2:
I checked it. This is good work! I created a new feature branch *usb-serial* and cherry-picked your commit 2dc296a there. I did some modifications:

* Renamed "userialN" device to "uttyN", because a similar name pattern is already used for the UART device
* Use DMA_BUFFER for rxData[] handed over to ControlMessage() (see [doc/dma-buffer-requirements.txt](https://github.com/rsta2/circle/blob/master/doc/dma-buffer-requirements.txt) )
* Check if m_pEndpointIn != 0
* Moved sample to addon/gpio/sample, because this is a very special sample, only very few people will have the RTK.GPIO device
* Implemented and used the USB device ID tables in usbdevicefactory.cpp
* Made the driver a little more quiet (log messages)
* Removed the hard-coded UART target for the log messages in the sample

Thanks so far!

I think, I will also provide a serial USB driver for the nice little BBC micro:bit computer, when you are ready with your drivers. It has a number of interesting sensors and can be connected using a USB CDC serial interface.

hinxx:
Thanks for accepting this PR, your comments and improvements!

Now that we have a baseline, I might add more functionality to the RTK.GPIO sample as we along. I'm also tempted to introduce hardware flow control.

Also, I'll try to add new USB to serial dongles based on other ICs in the coming days. I'll keep updating this PR, if that is OK with you.

rsta2:
I fully agree. Just update this PR and I will pick your commits for the *usb-serial* branch in this repo.

hinxx:
I'm not sure how to handle my `usb-serial` branch PR to your `devel` while (from now on) trying to work off the code you baselined in your `usb-serial branch. I just noticed was some weirdness going on on my side after trying to pull in your `usb-serial` branch.

rsta2:
Difficult to say from here, what the weirdness is and where is comes from. In any case, you need a branch as stable basis from where you can re-start now. I don't know, if you have this in your own repo. Otherwise you could also use the *usb-serial* branch in my Circle repo, which should be clean. Perhaps you have to create a new PR based on this branch or to *develop* branch and get rid of this PR, which seems to be broken. I will not do commits again for the USB serial feature, until you say, you are ready with it.

hinxx:
I'm in favor of creating a new PR where the next changes would on top of your newly created usb-serial branch.

rsta2:
That's probably the best idea. I revoked the link of this PR to issue #154, so that the issue remains open, when PR #155 is closed.

hinxx:
Closing in favor of new PR re-based on `usb-serial` branch.

PR #157, opened 2020-11-10
soundbasedevice: Allow fast path for same formats

dwhinham:
This is a trivial change that I hope makes sense!

The idea is that I wanted to be able to write into the queue of SoundBaseDevice as quickly as possible and avoid unnecessary conversion via `ConvertSoundFormat()`.

I have synthesizers that output floating-point samples, and in my kernel I convert these directly to 24-bit signed integers suitable for I2S DACs, and enqueue them.

With this change, the same "fast path" as defined for `SoundFormatSigned16` is generalized so that it applies whenever the write format matches the hardware format.


rsta2:
Yes, this makes sense! Merged it. Thanks!

PR #158, opened 2020-11-10
soundbasedevice: Allow queue allocation in frames

dwhinham:
Currently, `SoundBaseDevice` is limited because we can only allocate queues in terms of milliseconds.

For projects that want to have more precise control of the queue size (e.g. to reduce audio latency to an absolute minimum) it would be ideal to be able to allocate a sound queue in terms of frames. In my project, I am able to run reliably with the queue size equal to the chunk size (i.e. just one chunk in the queue), and a dedicated core rendering audio into the queue.

This PR may not be the cleanest way to do this as it just adds a new function and most of the code is just a copy of `AllocateQueueFrames()`, so if you can think of a better way to achieve this, I'd be more than happy to refine the PR! 

Thanks!

### DIFF
--- a/include/circle/soundbasedevice.h
+++ b/include/circle/soundbasedevice.h
@@ -84,6 +84,11 @@ public:
 	/// \note Not used, if GetChunk() is overloaded.
 	boolean AllocateQueue (unsigned nSizeMsecs);
 
+	/// \brief Allocate the queue used for Write()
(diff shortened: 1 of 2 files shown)


rsta2:
I think, it's not a problem, that the new method copies some code from `AllocateQueue()`, because the new method is small. Thus I merged it. Thanks!

PR #159, opened 2020-11-10
bcmpropertytags: Add GET_THROTTLED property tag

dwhinham:
Just a very simple PR; I currently have this defined in my own headers but thought it'd be nice to have it included with the others. 🙂

I'm using this tag to get the throttled status using a `TPropertyTagSimple`, as it's useful for detecting undervoltage conditions (user's PSU/USB cable is too weak) and throttle conditions (firmware has throttled the CPU as a result).

It might be a nice idea to have this functionality in Circle itself; maybe as an extension to the `CPUThrottle` class, so that a callback could be registered for when undervoltage/throttling is detected.


rsta2:
Merged it. Thanks! I have to think about, if further support in `CCPUThrottle` is possible and how this may look like.

dwhinham:
Thankyou for merging all three of my PRs so quickly! 🙂

rsta2:
Can you tell me, how the value returned by the GET_THROTTLED property mailbox call has to handled? I have seen in the Linux driver, that bit 16 is the under-voltage flag, but no clock throttled flag.

dwhinham:
Sure! The documentation for `vcgencmd` gives some clues here: https://www.raspberrypi.org/documentation/raspbian/applications/vcgencmd.md

I have some sample code here that may or may not be helpful: https://github.com/dwhinham/mt32-pi/blob/c18d922fa924347140772824f196d2b5ce0d80f0/src/power.cpp#L97

The upper bits of the value ("since power on" bits) are "sticky", and it appears that placing `0xFFFF` in the request clears them so you can detect new occurrences.

You can can also see how this is done in the Linux hwmon code: https://github.com/torvalds/linux/blob/master/drivers/hwmon/raspberrypi-hwmon.c

Hope this helps!

rsta2:
Yes, I think, this helps very much. Thank you!

rsta2:
There is an update to the `CCPUThrottle` class on the *develop* branch, which implements a system throttled callback function. You have to register a callback function using `RegisterSystemThrottledHandler()`, where `StateMask` is can be `SystemStateUnderVoltageOccurred | SystemStateThrottlingOccurred` for example. Then the application has to call `m_CPUThrottle.Update()` continuously. When the system throttled condition occurs, the callback function will be called with the current throttled state handed over.

dwhinham:
Perfect! Thankyou very much for implementing this - I can now remove some of my custom code and use your updated `CCPUThrottle` class.

Thanks for your hard work! 🙂

PR #167, opened 2020-11-28
Add option to allow I2S DAC to be master clock

dwhinham:
Hi!

With this patch, it becomes possible to use DACs that generate their own I2S clocks. The example I have been testing is the Blokas Pisound, which has an ADC (recording device) on-board, and seems to be wired such that the ADC provides the master clock.

Hence, to use this DAC, the Pi's I2S device must be configured to be a slave.

Apologies if the tab size is incorrect, I've tried to follow your indentation style - it seems to be correct if the tab size is 4. 🙂


rsta2:
Thanks! The tab size is 8. I will correct it, if necessary.

Issue #172, opened 2020-12-16
50MHz SD card clock speed


dwhinham:
Currently, `emmc.cpp` sets up the SD card interface with a fixed speed of 25MHz, and doesn't attempt to negotiate a faster rate of 50MHz either at 3.3V ("High Speed" mode) or 1.8V (SDR25 mode).

For Circle projects that do considerable amounts of file I/O, we can almost double the loading speed of larger files if we can switch to the higher speed mode. In [my project](/dwhinham/mt32-pi), users often have to wait for a large SoundFont file to load - some are only ~30MB, but others are >250MB, and so this has a big impact on loading times.

I've been measuring load times using a hack to change [this line](https://github.com/rsta2/circle/blob/42c0812385fe3b93c8f13b2410b4495c2d3fc385/addon/SDCard/emmc.cpp#L1703) to read `SwitchClockRate (base_clock, SD_CLOCK_HIGH);`, and it works very well for my modern SanDisk SD cards, but it would be better if we can check the SD card's capabilities and conditionally switch to the faster speed for better compatibility.

Would you be open to a patch that tries to implement this? I'm not an SD card expert but I could try to read some specs and attempt this - unless you think you would be able to add this more easily than me! 🙂

I also noticed that this definition for 1.8V mode is commented out:
https://github.com/rsta2/circle/blob/42c0812385fe3b93c8f13b2410b4495c2d3fc385/addon/SDCard/emmc.cpp#L69

Was there a reason for this, and does it still apply to modern Raspberry Pis?

Thanks for your consideration!

rsta2:
Reliable SD card access is not as easy to achieve, as it might look like. Circle uses a ported driver and one has to be very familiar with the whole technology to write such a driver. In Circle this driver must work reliable on RPi 1-4 and Zero with different interfaces (EMMC or SDHOST) and there are a lot of different SD cards out there, which should all work. I spent a lot of time to modify this driver for Circle, for different RPi models and requirements, but I wouldn't say I'm familiar with the SD card technology. That's why I was happy, that the existing driver seems to work reliable, which was the most important requirement for me.

I understand that for your application purpose, SD card speed is important. So if you have the time to dive into SD card technology, so that you can modify the driver and can tell you have a good feeling that it will work with your modifications on any RPi model and on (say) most SD cards, I would be glad to accept these modifications for Circle.

Please note:

[...]

PR #174, opened 2020-12-28
addon/SDCard: Enable optional High Speed/SDR25 mode

dwhinham:
As originally discussed in #172.

This commit enables an optional `SD_HIGH_SPEED` definition that, if enabled, will cause the EMMC driver to try to negotiate a higher clock speed. This should work for all combinations of EMMC/SDHOST driver configurations.

This is done by sending CMD6 in Mode 0 to read the supported access modes from the SD card. If the High Speed/SDR25 bit is enabled, call CMD6 again in Mode 1 to request that the SD card switch to this mode, and then switch the host clock rate to match if successful.

CMD6 is only supported by SD cards that support the Physical Layer Specification version 1.10 or above, so this procedure is wrapped in a version check (reading the version number that was received from the SCR register).

The [Linux driver](https://elixir.bootlin.com/linux/latest/source/drivers/mmc/core/sd.c#L1113) and the [Physical Layer Simplified Specification](https://www.sdcard.org/downloads/pls/click.php?p=Part1_Physical_Layer_Simplified_Specification_Ver8.00.jpg&f=Part1_Physical_Layer_Simplified_Specification_Ver8.00.pdf&e=EN_SS1_8) were used as references for implementing this behaviour.

1.8V mode operation has not been considered or tested, though if 1.8V mode was enabled for Pi 4 in a future commit, this code shouldn't need any changes - the clock speed change mechanism is pretty much the same, but it's called "SDR25" rather than "High Speed".

---

Not part of this PR is a simple test application that was used to test performance and reliability of this patch. Code for the test app is available here: https://github.com/dwhinham/circle/tree/sdcard-test

The test is not exhaustive or perfect, but gives a pretty good indication of stability. A variety of old and new SD cards were tested against Pi Zero W, Pi Model B, Pi 3A+, and Pi 4B. 

So far I have encountered no failures, which is reassuring. The code may even be safe to enable permanently as it's just doing a very similar thing to what Linux does, but perhaps more testing would be a good idea to confirm this.

Test description:

1. A 256MB file full of random data is created on a PC, using `dd if=/dev/urandom of=testfile.bin bs=1M count=256  status=progress`. This file is copied to the root of the SD card.
2. The SHA256 sum of the file is taken and noted: `sha256sum testfile.bin`
3. A kernel is compiled for the target Pi, first with `SD_HIGH_SPEED` undefined, and then with it defined (e.g. by adding `DEFINE += -DSD_HIGH_SPEED` into `Config.mk`).
4. The Pi runs the kernel, which then loads the entire 256MB file into memory, and performs a SHA256 hash on the data.
5. The kernel then writes the file from memory back to a new file, `testfile2.bin`.
6. The user verifies that the SHA256 hash was correct when the Pi read the file (by reading the kernel's log), and that the SHA256 hash of the `testfile2.bin` that was written by the Pi matches the original (by using a PC).

Test results:
image

Editable spreadsheet:
https://docs.google.com/spreadsheets/d/1nBtIlqOCRY-ZC1rlxMgWXzXILBHJniGg2TMiyOLvr70/edit?usp=sharing

Note that from the results, even with `SD_HIGH_SPEED` defined, cards that do not support High Speed mode (such as the SanDisk microSD 1GB) simply operate as normal at Default Speed.

I've attached binaries of the test app for convenience here with 32-bit kernels for Pi 0/1, 3, and 4:
[sdcard-test.zip](https://github.com/rsta2/circle/files/5745442/sdcard-test.zip)

All you have to do is create the test file (as described in step 1) and rename one of the kernel images. `_SLOW` kernels were compiled with `SD_HIGH_SPEED` **undefined**, `_FAST` kernels were compiled with `SD_HIGH_SPEED` **defined**.

I hope this makes sense and I'd be glad to make any corrections or respond to feedback.
Thanks for your consideration! 🙂

### DIFF
--- a/addon/SDCard/emmc.cpp
+++ b/addon/SDCard/emmc.cpp
@@ -68,6 +68,9 @@
 // Enable 1.8V support
 //#define SD_1_8V_SUPPORT
 
+// Enable High Speed/SDR25 mode
+//#define SD_HIGH_SPEED
+
 // Enable 4-bit support
 #define SD_4BIT_DATA
 
@@ -277,7 +280,7 @@ const u32 CEMMCDevice::sd_commands[] =
 	SD_CMD_INDEX(3) | SD_RESP_R6,
 	SD_CMD_INDEX(4),
 	SD_CMD_INDEX(5) | SD_RESP_R4,
-	SD_CMD_INDEX(6) | SD_RESP_R1,
+	SD_CMD_INDEX(6) | SD_RESP_R1 | SD_DATA_READ,
 	SD_CMD_INDEX(7) | SD_RESP_R1b,
 	SD_CMD_INDEX(8) | SD_RESP_R7,
 	SD_CMD_INDEX(9) | SD_RESP_R2,
@@ -1511,6 +1514,7 @@ int CEMMCDevice::CardReset (void)
 	m_device_id[3] = 0;
 
 	m_card_supports_sdhc = 0;
+	m_card_supports_hs = 0;
 	m_card_supports_18v = 0;
 	m_card_ocr = 0;
 	m_card_rca = 0;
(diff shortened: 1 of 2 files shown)


rsta2:
Some more notes now. I tested your update to the SD card driver with your test program according to your specification on a RPi 1B, 2B, 3B and 400. I only had one SD card and one microSD card available for this test. Everything was working well. The results are in the spreadsheet (lines 24-28). Even that my SD cards are not the of the newest spec. the increase especially in read performance was remarkable.

There is an effect in the performance on the RPi 400. The measured times with it are greater than the times for the RPi 1-3 with the same microSD card. I guess, this is still caused by the EMMC driver, which is not that fast as the SDHOST driver, which was directly ported from Linux.

Having a look at the all results I think, with the right microSD card, your update can be a great improvement especially for the times needed to load, greater amounts of data. So we should include this update into Circle. Your added code is localized to a few lines and makes absolutely sense. Perhaps we can enable the `SD_HIGH_SPEED` define by default. What do you think?

dwhinham:
Thankyou for taking the time to run the tests and add the results to the spreadsheet!

Your findings are very similar to mine - even older cards on older Raspberry Pi models can benefit from a measurable speed increase, so these changes should hopefully be useful for all users of Circle.

You also noticed the same thing with your Pi 400 as I did with my Pi 4 B - there is a speed increase, but not as much as the models using SDHOST - maybe there is some future work that can be done to enhance the EMMC driver further. The Pi 4's controller can also use DDR50 speed, but this requires 1.8V signalling and the slightly more complex "Ultra High Speed" initialization sequence, so this could also be a future task.

For now, I agree - I think this is a nice "low-hanging fruit" that doesn't need many changes - I think I'd be confident enough in it to enable `SD_HIGH_SPEED` by default.

We could do something like:
```c++
#ifndef NO_SD_HIGH_SPEED
#define SD_HIGH_SPEED
#endif
```

...in `sysconfig.h` to enable it by default, but offer a way to disable it, similar to the define for `NO_SDHOST`, if you think that'd be a good idea? Either way, as a user of Circle I'd be happy with any of these options. 🙂

rsta2:
Ok, I added the `SD_HIGH_SPEED` option to *sysconfig.h* as suggested and merged everything to the *develop* branch. Thank you again for this contribution to Circle and for testing everything that carefully!

dwhinham:
Great! Thanks for giving this PR your attention so quickly and also for adding me to `CREDITS` - much appreciated!

PR #177, opened 2021-01-05
Added some tips and notes to the bootloader instructions

toptensoftware:


toptensoftware:
First time user of circle... added some notes and tips that should make it a little easier for others.

rsta2:
I have added the information on using WSL to the *develop* branch. Thanks! I left the other notes out, because I thought they were doubled, obvious or too special. Also I want to keep the main README short, because it is often read.

PR #178, opened 2021-01-06
Fix the USB standard gamepad report decoding.

Tom1975:
Hi,

While using the great circle lib for my personnal project (Amstrad CPC emulator for bare-metal pi3/4), I encountered a problem using my gamepads : 
A kind of Generic gamepad was not working as expected. after some investigation, I found that the report decoding is not correct, when the bit size of a field is not a multiple of 8 (I have 4 and 12 bit size, here).
After some more investigation, I found a similar function, working correctly with the HID mouse. So, I suggest to use the same function for both.
It's the purpose of this pool request.

Kind regards,

### DIFF
--- a/include/circle/usb/usbgamepadstandard.h
+++ b/include/circle/usb/usbgamepadstandard.h
@@ -38,10 +38,6 @@ protected:
 	void DecodeReport (const u8 *pReportBuffer);
 
 private:
-	static u32 BitGetUnsigned (const void *buffer, u32 offset, u32 length);
-	static s32 BitGetSigned (const void *buffer, u32 offset, u32 length);
-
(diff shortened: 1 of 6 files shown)


rsta2:
Thanks for the pull request! I didn't merge it directly, because I wanted to keep HID report extracting code out of the class CUSBHIDDevice. There is a HID parser in CUSBGamePadStandardDevice and in CUSBMouseDevice, but not in CUSBHIDDevice. So I simply copied the correct code from CUSBMouseDevice to CUSBGamePadStandardDevice. The doubled code should not be a problem for these small functions. The update is on the *develop* branch at the moment.

PR #181, opened 2021-01-07
Moved keyboard repeat enable and rate settings to sysconfig.h

toptensoftware:


rsta2:
Thanks. Unfortunately, what you cannot know, I had problems with the keyboard repeat function earlier, especially on the RPi 1. That's the reason why this function is not enabled by default and I don't want to expose it too much by placing it in sysconfig.h. The function can still be enabled by adding `DEFINE += -DREPEAT_ENABLE` to *Config.mk*, if it is really wanted.

toptensoftware:
ok no problem.. thanks for the tip about -DREPEAT_ENABLE, that'll suffice nicely.

PR #184, opened 2021-01-11
Fix for CTask::WaitForTermination when task has been deleted by scheduler

toptensoftware:
This is a fix for a condition where `CTask::WaitForTermination` might be called on a task that's been deleted.  Because the scheduler deletes tasks when they're terminated its difficult to tell if a task pointer is still valid before waiting for its termination.

Consider this example:

```cpp
class CMyTask : public CTask
{
public:
	CMyTask() 
	{
		printk("Task constructor");
	}
	virtual ~CMyTask() 
	{
		printk("Task destructor     <--- Task ptr no longer valid");
	}
	virtual void Run() override
	{
		printk("Task running...");
		CScheduler::Get()->MsSleep(500);
		printk("Tash finished.");
	}
};


TShutdownMode CKernel::Run (void)
{
	m_Logger.Write (FromKernel, LogNotice, "Compile time: " __DATE__ " " __TIME__);

	printk("Creating Task");
	CTask* pTask = new CMyTask();

	printk("Main task sleeping for a bit...");
	CScheduler::Get()->MsSleep(1000);

	printk("Waiting Task        <--- Task ptr better still be valid");
	pTask->WaitForTermination();

	printk("Finished");
	return ShutdownReboot;
}
```

which produces the following output:

```
00:00:00.67 printk: Creating Task
00:00:00.67 printk: Task constructor
00:00:00.68 printk: Main task sleeping for a bit...
00:00:00.68 printk: Task running...
00:00:01.18 printk: Tash finished.
00:00:01.19 printk: Task destructor     <--- Task ptr no longer valid
00:00:01.68 printk: Waiting Task        <--- Task ptr better still be valid
00:00:01.69 printk: Finished
```

At the moment it works out pure luck that the memory where the task was hasn't been re-used.

The attached fix works by updating `WaitForTermination` to validate its `this` pointer by checking it's still registered in the scheduled tasks list.  Doing this before accessing any member variables should prevent crashes.

It also provides a public scheduler method `IsValidTask` that might be useful for other situations where the validity of a task pointer needs to be determined.

### DIFF
--- a/include/circle/sched/scheduler.h
+++ b/include/circle/sched/scheduler.h
@@ -39,6 +39,7 @@ public:
 	void usSleep (unsigned nMicroSeconds);
 
 	CTask *GetCurrentTask (void);
+	boolean IsValidTask(CTask* pTask);
 
 	void RegisterTaskSwitchHandler (TSchedulerTaskHandler *pHandler);
 	void RegisterTaskTerminationHandler (TSchedulerTaskHandler *pHandler);
@@ -49,6 +50,7 @@ public:
 	{
 		return s_pThis != 0 ? TRUE : FALSE;
 	}
+
 
 private:
 	void AddTask (CTask *pTask);
--- a/lib/sched/scheduler.cpp
+++ b/lib/sched/scheduler.cpp
@@ -121,6 +121,18 @@ CTask *CScheduler::GetCurrentTask (void)
 	return m_pCurrent;
 }
 
+boolean CScheduler::IsValidTask(CTask* pTask)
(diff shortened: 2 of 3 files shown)


rsta2:
This method was only used in the pthread emulation code and was protected by a test [there](https://github.com/rsta2/circle/blob/master/addon/linux/pthread.cpp#L110), if the task has already been terminated. So I don't think, there was really a bug before, but with your modification it can be used for different purposes now and the pthread emulation should still work. Thanks!

PR #186, opened 2021-01-13
Fix for inability to separate raw key events from multiple keyboards

toptensoftware:
This PR adds a user supplied callback argument to the `CUSBKeyboardDevice::RegisterKeyStatusHandlerRaw`.

Since key transitions are determined by comparing one set of RawKeys with a previous set it's important that the set being compared to is from the same device. Without this fix, a client can't separate out the RawKeys from multiple keyboards making it impossible to determine key state transitions as the RawKeys for each device will be interleaved.

By adding a user supplied callback argument the client can match the received callback events to a particular device.

The same thing applies for raw mouse handler button states.

Unfortunately this is a breaking API change but solves an otherwise impossible to rectify issue.  I've updated all the samples that use these methods.

### DIFF
--- a/addon/vc4/interface/sample/hello_triangle2/kernel.cpp
+++ b/addon/vc4/interface/sample/hello_triangle2/kernel.cpp
@@ -21,7 +21,7 @@
 
 extern "C" int _main (void);
 extern "C" void mouse_callback (unsigned nButtons, int nDisplacementX, int nDisplacementY,
-				int nWheelMove);
+				int nWheelMove, void* arg);
 
 static const char FromKernel[] = "kernel";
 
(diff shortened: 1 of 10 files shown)


rsta2:
It is possible to install different callback functions for different keyboard or mouse instances. So if you really want to support multiple instances of the keyboard or mouse, you can have multiple stubs and register these as raw event handler. From these stubs you can branch to your actual raw event handler. Because this will be needed only in very few cases (most applications use only one instance of these devices) and it is a API breaking change, I cannot accept this.

toptensoftware:
Understood, I'll just keep the change in my fork.

PR #187, opened 2021-01-13
Improvements to Task Synchronization primitives

toptensoftware:
This PR adds:

* A new method `CSynchonizationEvent::WaitWithTimeout()` that yields the current task until either the event is signaled or a timeout period expires.
* Support for multiple tasks waiting on a single `CSynchronizationEvent`
* A new class `CSynchronizationMutex` that provides mutual exclusion capability across yielding tasks

It also includes a new sample program `41-synchronization` that demonstrates/exercises all of the above and I've added CSynchronizationMutex to the doc/classes.txt file.

This is a non-breaking API change.

### DIFF
--- a/doc/classes.txt
+++ b/doc/classes.txt
@@ -146,6 +146,7 @@ Scheduler library
 * CTask: Overload this class, define the Run() method to implement your own task and call new on it to start it.
 * CScheduler: Cooperative non-preemtive scheduler which controls which task runs at a time.
 * CSynchronizationEvent: Provides a method to synchronize the execution of a task with an event.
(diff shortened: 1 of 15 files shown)


rsta2:
The PR has been merged. I did some minor modifications to it afterwards. I think the added features are self explanatory and do not need a separate sample. That's why I removed it from the repo, but kept it locally for testing. Thank you for this valuable contribution!

toptensoftware:
I agree the sample's not really necessary but wanted to include a way to test the changes with the PR. Thanks for taking this PR.

PR #188, opened 2021-01-16
New Boot Loader Flash Tool

toptensoftware:
This PR includes a new bootloader flasher script that improves flash operations in a couple of ways:

* Pings the bootloader and waits for ack that it's ready before sending
* Waits for ack after sending the file instead of assuming it succeeded
* Can reset the bootloader from a failed or canceled previous load
* Can send magic reboot strings
* Can start monitoring after the flash (including switching baud rates if necessary)
* Allows flashing from inside WSL2 machines (that don't support serial ports)

In addition to the the flash script itself, there's also:

* A slight tweak to the bootloader for ping/ack/reset
* An updated version of Rules.mk to optionally enable the new flasher and to support WSL
* Updated documentation on the bootloader describing how to enable it
* Updated windows build documentation explaining how to use it under WSL 2

The new script also benefits from being written in NodeJS instead of Python:

1. I couldn't get python3.exe to launch from inside a WSL machine.  
2. Python on some Windows 10 machines can be frustratingly slow to start - 5 to 10 seconds.

I don't know why python seems to have these issues, but I couldn't find solutions to either and Node doesn't seem to have these problem.

I consider all these changes low risk since everything is optional.  The Rules.mk file is only affected if enabled in configuration.  The change to the bootloader is minimal.

### DIFF
--- a/Rules.mk
+++ b/Rules.mk
@@ -190,11 +190,49 @@ $(TARGET).hex: $(TARGET).img
 	@echo "  COPY  $(TARGET).hex"
 	@$(PREFIX)objcopy $(TARGET).elf -O ihex $(TARGET).hex
 
+# Command line to run node and python.  
+# Including the '.exe' forces WSL to run the Windows host version
+# of these commands.  If putty and node are available on the windows 
+# machine we can get around WSL's lack of serial port support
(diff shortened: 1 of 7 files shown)


toptensoftware:
Please ignore. Closing this pull request... will issue a new one with squashed commits and more improvements.

PR #194, opened 2021-01-21
New Bootloader and Flasher Script

toptensoftware:
This PR includes a new bootloader and flasher script that improves flash operations in a couple of ways:

* Fast flash mode is twice as fast
* Pings the bootloader and waits for ack that it's ready before sending
* Waits for ack after sending the file instead of assuming it succeeded
* Transfers are now using the checksums in the .hex file to validate the transfer
* The flash tool can reset the bootloader from a failed or canceled previous load
* The flash tool can send magic reboot strings 
* The flash tool can automatically start monitoring the serial port after the flash (including switching baud rates if necessary) (receive only)
* Configurable "go delay" can cause the bootloader stall between receiving the go command and actually launching the flashed program (can be used if missing the start of the serial log)
* Allows flashing from inside WSL 2 machines even though they don't support serial ports.

In addition to the the bootloader and flash script, there's also:

* An updated version of Rules.mk to optionally enable the new flasher and to support WSL
* Updated documentation on the bootloader describing how to enable it
* Updated windows build documentation explaining how to use it under WSL 2

The new script also benefits from being written in NodeJS instead of Python. I couldn't get python3.exe to launch from inside a WSL machine. Also python3 on some Windows 10 machines can be frustratingly slow to start - 5 to 10 seconds, so this is an alternative if suffering from this issue.

There are a fair few changes here but I think they're mostly low risk. Nothing touches the core circle library and most changes only take effect when enabled. The biggest change is to the bootloader which has been significantly reworked to handle fast flash mode (and other new features)

PS: Sorry it took a few goes to get this PR right.  The previous cancelled PRs worked at slow baud rates, but I found the Pi couldn't keep up when testing at higher rates.  The fix required a complete rework of the bootloader but in testing I've now been able to reliably fast flash a Pi2 @ 2M and Pi3 @ 3M.  Previously I could get only get those rates in the old non-fast ascii mode.

### DIFF
--- a/Rules.mk
+++ b/Rules.mk
@@ -190,11 +190,50 @@ $(TARGET).hex: $(TARGET).img
 	@echo "  COPY  $(TARGET).hex"
 	@$(PREFIX)objcopy $(TARGET).elf -O ihex $(TARGET).hex
 
+# Command line to run node and python.  
+# Including the '.exe' forces WSL to run the Windows host version
+# of these commands.  If putty and node are available on the windows 
+# machine we can get around WSL's lack of serial port support
+ifeq ($(strip $(WSL_DISTRO_NAME)),)
+NODE=node 
+PUTTY=putty
+PUTTYSERIALPORT=$(SERIALPORT)
+else
+NODE=node.exe
+PUTTY=putty.exe
+PUTTYSERIALPORT=$(subst /dev/ttyS,COM,$(SERIALPORT))		# Remap to windows name
+endif
+
+ifeq ($(strip $(USEFLASHY)),)
+
+# Flash with python
 flash: $(TARGET).hex
 ifneq ($(strip $(REBOOTMAGIC)),)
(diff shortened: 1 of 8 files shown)


rsta2:
I merged it. Great tool! First I had some doubts, because of using nodejs instead of python, but because it's compatible with the previous solution, this is not a problem. Thank you.

toptensoftware:
Hi Rene,  thanks for merging.  

I've been using this fairly intensely for a couple of weeks now.  99% of the time it works perfectly.  

There's one small issue where very occasionally it'll not see the bootloader ack response from the go command and it'll sit and wait indefinitely.  I'm still trying to nail down exactly what's going on and will post a fix if/when I find it.  In the meantime, just Ctrl+C and run it again and it again to recover.

If you notice any other issues, let me know.

rsta2:
OK, understood.

rsta2:
I modified *tools/flasher.py* so that it sends a reset command ("R") before sending the kernel image. It did not work on the Raspberry Pi Zero without it any more.

PR #196, opened 2021-01-21
Added GCC compile time checks for CString::Format and CLogger::Write printf style arguments

toptensoftware:
There are other places that this could be used, but these seemed the most important two.

### DIFF
--- a/include/circle/logger.h
+++ b/include/circle/logger.h
@@ -55,7 +55,8 @@ public:
 
 	void SetNewTarget (CDevice *pTarget);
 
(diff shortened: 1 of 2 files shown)


rsta2:
Unfortunately there is a problem, when compiling the submodule *addon/wlan/hostap/* after applying this change. One gets some ugly warnings, because the word `printf` has been defined as a macro, so that the function printf() will get a different name in this module. Otherwise there would be multiple instances of printf(), when using the circle-stdlib project.

toptensoftware:
I thought I built everything to make sure it didn't affect anything but must have missed that project.  No problem

PR #212, opened 2021-03-17
PWM audio: allow using gpio18/19 for output.

flabbergast:
This PR adds two more #defines which allow moving pwm audio output pins from 12/13 to 18/19 (individually).
Arguably this is only useful on a Pi Zero (the others have a hardwired jack), but I have a device that uses those (namely a [GPi Case](http://retroflag.com/GPi-CASE.html) I mentioned in another Issue).
I tested it with your pwm audio sample code and it works as expected.
Feel free to reject if you feel it just adds unnecessary clutter to sysconfig.h.

### DIFF
--- a/include/circle/sysconfig.h
+++ b/include/circle/sysconfig.h
@@ -104,12 +104,21 @@
 #define GPU_L2_CACHE_ENABLED
 #endif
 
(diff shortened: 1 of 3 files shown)


rsta2:

PR #219, opened 2021-04-20
Added Ltd FT2232C/D/H Dual UART/FIFO IC to list of ft231x devices

hpingel:
Linux reports the device as:
Bus 002 Device 005: ID 0403:6010 Future Technology Devices International, Ltd FT2232C/D/H Dual UART/FIFO IC

One of the Sidekick64 network testers has this device and I patched it into to the source. He reports it to be working fine. Would be cool if you could at this to circle, Rene! Thank you in advance!


rsta2:

PR #225, opened 2021-05-04
Complete TODO item of ANSI reverse code, on way to ANSI colour support.

Xalior:
Hi there, this is an "experimental commit" to to see if you approve of the "BackgroundColor" property in screen.cpp.

This is a partial port of my own private repo with full 16colour ANSI support, Fore and Background.  But mostly I want to check the style meets your approval.

I've chosen to do this facet first as I noticed you'd marked it as a TODO item in the source code.

If this doesn't meet with your approval I'd sorely like a tip to get it to "pass muster", so I can change the style, and then upgrade the rest of my code to match.  I've also checked this with `circle-stdlib` and `printf` et al is working fine too.

I'll submit a pull request to `circle-stdlib` with a full colour demo, once I've got this all merged into your main repo here.

I've also, as a second commit in this pull request, modified the 03-screentext demo to inverse the actual characters in the grid, as an example of using the control code for inverse.  You, obviously, don't have to take any of these - but the second commit (the demo one) is "extra optional".

Looking forward to your response, before I continue, and send you a pull request for the remainder of the 16colour work.
Hopefully you accept, but - of course - feel free to do with as you wish. Any "stylistic" edits/input you make I will do my best to adhere to on further pull requests.

-Dx

### DIFF
--- a/doc/screen.txt
+++ b/doc/screen.txt
@@ -15,6 +15,7 @@ ESCAPE SEQUENCES
 \E[0m		End of bold or half bright mode
 \E[1m		Start bold mode
 \E[2m		Start half bright mode
+\E[7m		Start reverse video mode
 ^I		Move to next hardware tab
 \E[?25h		Normal cursor visible
 \E[?25l		Cursor invisible
--- a/doc/screen.txt
+++ b/doc/screen.txt
@@ -16,6 +16,7 @@ ESCAPE SEQUENCES
 \E[1m		Start bold mode
 \E[2m		Start half bright mode
 \E[7m		Start reverse video mode
+\E[27m		End reverse video mode
 ^I		Move to next hardware tab
 \E[?25h		Normal cursor visible
 \E[?25l		Cursor invisible
--- a/include/circle/screen.h
+++ b/include/circle/screen.h
@@ -83,6 +83,7 @@ struct TScreenStatus
 	unsigned	nCursorY;
 	boolean		bCursorOn;
 	TScreenColor	Color;
+	TScreenColor	BackgroundColor;
 	boolean		bInsertOn;
 	unsigned	nParam1;
 	unsigned	nParam2;
@@ -136,6 +137,7 @@ public:
 	/// \param nPosX X-Position of the pixel (based on 0)
 	/// \param nPosY Y-Position of the pixel (based on 0)
 	/// \param Color The color to be set (value depends on screen DEPTH)
+	/// \param BackgroundColor The color to be set (value depends on screen DEPTH)
 	void SetPixel (unsigned nPosX, unsigned nPosY, TScreenColor Color);
 	/// \brief Get the color value of a pixel
 	/// \param nPosX X-Position of the pixel (based on 0)
@@ -206,6 +208,7 @@ private:
 	unsigned	 m_nCursorY;
 	boolean		 m_bCursorOn;
 	TScreenColor	 m_Color;
+	TScreenColor	 m_BackgroundColor;
 	boolean		 m_bInsertOn;
 	unsigned	 m_nParam1;
 	unsigned	 m_nParam2;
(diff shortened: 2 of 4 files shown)


Xalior:
image

rsta2:
Hi, to bring more color onto the Circle screen is generally a great idea! Thanks for that and for the pull request.

Talking about the reverse mode, I think, we have to agree about, how this should exactly work. When implementing the `CScreenDevice` class I had the Linux console in mind and I tested the behavior of the `\x1b[7m` sequence there now. There are two differences:

1. The `\x1b[7m` sequence starts the reverse mode, but a second `\x1b[7m` does not stop it. One has to send `\x1b[0m` to stop the reverse mode.

2. When the reverse mode is on, and you send a cursor home and clear screen sequence, that does not clear the whole background to white. The reverse mode does only apply to the written characters, not to the entire screen.

So if you could update your pull request that way, it would be great. I think, you adapted my coding style (use tabs, tabs are 8 characters wide, maximum line length is 100 characters, braces "{}" should be on a separate code line) well.

Xalior:
No worries! I was referenceing an old VTclone on my desk, so a considerably older implimentation, but knowing what you're "aiming for" gives me a better reference platform.  

I've had a "flipping nightmare" getting VSCode not to mangle your source, so I think I'll just put together a pull request in Vi next time.  At least that is less obnoxiously opinionated in the default setup! :) )

I'll add another attribute to the Screen class for Reverse Video - since otherwise clear screen, etc won't know which background is the background :)

Will push that, and retag you in this thread once done. Cheers for the tips, and positive encouragement <3 

-Dx

Xalior:
Second Atttempt ;-)   I've used `man console_codes` as my guide here.

1) Screen.cpp updated with two new helpers:
   GetTextColor (), and GetTextBackgroundColor () -- these are to just to make it easier to work out the plotting colours, without having to constantly perform ternary operations on the m_ReverseAttribute.
2) Said m_ReverseAttribute, engaged and disabled as mentioned in your comment re: 0/27.
3) Helpers only used for actual text output, other "clearing" modes still reference the actual set background colour explicitly.
4) I've left the m_BackgroundColor parameter in, since while currently it's always COLOR_BLACK, it'll be required for full ANSI colour support post this PR.
5) Example control codes updated accordingly.

Programmatic erase, clear, etc. use the m_BackgroundColor attribute, still, as I could not find a clear definition in `console_codes` manpage to the "specific" behavior, so based it upon your clearing comments above, about written characters.

Hopefully that's more in line with your intentions @rsta2!

-Dx

rsta2:
Thanks for updating your pull request. This looks good now. There is one small thing left: When the system option `SCREEN_HEADLESS` is defined, there is a problem, because the member `m_BackgroundColor` is not defined then [here](https://github.com/Xalior/circle/blob/feature/ANSIinverse/lib/screen.cpp#L938). I think, `BLACK_COLOR` should be returned there as before, because we do not need to handle color with this option defined. If you would correct this in your PR, I will merge it then. You can also add the new `\E[7m` sequence to *doc/screen.txt*, if you want to. Thanks!

Xalior:
It is so! 👍 

Shall now branch from here, and will begin pestering you - anew - when colours are ready for your watchful eye! 👀  

-Dx

rsta2:
Merged it, thanks. There will be a new release in May and I think, the ANSI color support cannot be part of it yet. So there's no hurry for it. For the new pull request it is probably better to branch of from Circle's *develop* branch again.

Xalior:
> There will be a new release in May and I think, the ANSI color support cannot be part of it yet. 
If you don't mind me asking, what's the reasoning behind that?

> For the new pull request it is probably better to branch of from Circle's develop branch again.
Was exactly my plan, and why I named this one specifically reverse :)

-Dx

rsta2:
The reason is, that I closed the new feature list for the May release now. I'm still waiting for a clarification regarding one feature, but despite of this I will only do bug fixes and update the documentation before the release. Then I have to do the release tests, which will take some days. You can branch off from develop at any time and develop the new feature, but it will be part of the following release. Experience tells, that integrating a new feature takes time and I want to concentrate on preparing the upcoming release now.

Xalior:
Perfectly logical - thanks for the clarification <3

PR #228, opened 2021-05-29
RFC feature/ANSIcolours (a request for comment PR)

Xalior:
Hello again, and congratuations on Step 44 🥇 

Here's my first pass at the ANSIcolour support we discussed earlier in the month - I've tagged this RFC because I am "concerned" about the COLOR16 macro it seems to create more "washed out" colours than I would expect - and am wondering if COLOR8 works, since your code in `develop` uses the COLOR16 macro.

Therefore I'd appriciate your input, if possible, when you have the spare brainspace, and time!

This is the 32bit mode -- the "bright yellow" and "bright green" are the level of "vibrancy" I would expect.
COLOR32

This is the 16bit mode -- and the colours appear more "washed out"
COLOR16

As per usual, thanks for your hard work, your time, and your help on the matter!

-Dx


Xalior:
(there is also a bug in the cursor redraw - as seen in the pictures, but that for after the colourmaths! ;-) )

rsta2:
Thank you for the Pull Request and for appreciating the new Circle release! Please give me two days to check this.

rsta2:
I had a look at your code. Some hints:

* For the color definitions I would use the VGA colors from [here](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors). That are values for the `COLOR32()` definitions. For `COLOR16()` you have to divide each value by 8. Unfortunately I don't know, why the colors look "washed out", but perhaps with the VGA colors from this website this effect will go away?

* The `COLOR8()` definitions will not work that way. 8-bit colors are indices between 0 and 255 into a color palette, which is defined [here](https://github.com/rsta2/circle/blob/master/lib/screen.cpp#L86) for the currently available 3 colors (NORMAL, HIGH and HALF). BLACK is the default for all other indices and is not explicitly set here. I would suggest to define the colors from the VGA palette for the color indices 1-15 and and give NORMAL_COLOR the value of the white index, HIGH_COLOR of the red index and HALF_COLOR of the blue index.

* That the cursor symbol is still shown in the image is a problem. I thought about this and I think, you can only save the pixels at the cursor position into an internal array in `CScreenDevice` and then display the cursor. When the cursor is removed from a screen position, the pixels from the array are copied to the screen, so that the cursor disappears. I guess, the previous way to simply invert the pixels at the cursor position will not work any more.

In any way it would be good to have the ANSI colors on the Circle screen with these modifications.

rsta2:
How is the state of your work? Please let me know, if you need support.

Xalior:
I will, just some prior commitments (work, and a kickstarter campain I am helping to deliver) eating my time -- and got sidetracked for some time with post-vaccine side effect.

I promise to bug you if I need help! :)

Xalior:
Had a couple of free hrs, so confirmed the colours matched Wikipedia ones, and switched 16bit to a "quick hack" via bitshifting to confirm they looked good.  They do.  While the bitshifting won't add anything to runtime (they're compile time optimised out, or -- should be) they look gross in the codebase, so will swap these for hard-coded values before the final merge.

This update is purely just so you know what's happened - not ready to merge/review yet, but wanted to update you to progress as is.

-Dx

rsta2:
I will collect your commits, when you have finished this feature and have a look at it. Please don't merge other commits/branches on this feature branch, if they are not needed to develop the feature! It will be difficult to extract the commits needed for the feature then in the end. Also the commit message is normally not the place for todo lists and other info, which is not related to the commit.

Xalior:
That's 8, 16 and 32bit "real time" colour mode done - all based on the VGA palette from wikipedia.  And sample 42 which is a simple colour chart, and "repeatedly moving cursor" demo in sample 42.

rsta2:
I merged your modifications for ANSI color support in one commit and some small changes from myself to the *develop* branch. Your sample program has been moved to the test/ subdirectory, because I think, it is more important for testing than for given usage info to the user. Thank you for this contribution.

PR #230, opened 2021-05-31
addon/SDCard: don't check the result of GetTag so it can work on QEMU

richardchien:
According to [qemu/hw/misc/bcm2835_property.c](https://gitlab.com/qemu-project/qemu/-/blob/master/hw/misc/bcm2835_property.c), QEMU doesn't implemented the property tag `PROPTAG_SET_SDHOST_CLOCK`, but it does implement an SDHost. Remove the code that checks the result of `GetTag` can make SDHost driver work on QEMU, like [what `raspberrypi/linux` does](https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/mmc/host/bcm2835-sdhost.c#L2127).

### DIFF
--- a/addon/SDCard/sdhost.cpp
+++ b/addon/SDCard/sdhost.cpp
@@ -262,10 +262,7 @@ int CSDHOSTDevice::set_sdhost_clock (u32 msg[3])
 	memcpy (SetSDHOSTClock.msg, msg, sizeof *msg);
 
 	CBcmPropertyTags Tags;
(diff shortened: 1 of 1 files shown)


richardchien:
A screenshot to prove it works😃:

截屏2021-05-31 22 56 13

rsta2:
I cherry-picked your commit on the *develop* branch, where development takes place. I was not aware, that the SDHOST is supported by QEMU and thought, the SD card works only with the EMMC interface there. Great, that it works with SDHOST now! Thank you for this contribution.

richardchien:
You're welcome!

PR #232, opened 2021-06-08
FIXED:pwmoutput value of ARM_PWM_CTL_MSEN2 is incorrect

mitsuji:
Hi! Thank you for the wonderful library.

PWM_CHANNEL2 seems not working well. I tried to fix it.
https://www.raspberrypi.org/app/uploads/2012/02/BCM2835-ARM-Peripherals.pdf
Page 142


rsta2:
Indeed this bit setting was wrong. Thank you for the correction and for appreciating Circle!

hpingel:
Does this issue also need to be fixed in file pwmsoundbasedevice.cpp? There is another define in this file:
#define ARM_PWM_CTL_MSEN2	(1 << 14)

rsta2:
This define is not used here. I fixed it anyway. Thanks.

PR #233, opened 2021-06-13
Always supply memmove() in Circle

smuehlst:


rsta2:
Thank you. I will prepare a hotfix release tomorrow and merge your commit then. There is another issue, which needs to be fixed already.

rsta2:
The hotfix release 44.1 is available, which includes your patch.

PR #242, opened 2021-08-06
sdhost: correct gpio mode for pin 34-39

richardchien:
In `addon/wlan/ether4330.c`, the GPIO setting seems to be correct:

https://github.com/rsta2/circle/blob/da58255b60378ce8ee329ca66c2861d82e732562/addon/wlan/ether4330.c#L485-L495

But in `addon/SDCard/sdhost.cpp`:

https://github.com/rsta2/circle/blob/da58255b60378ce8ee329ca66c2861d82e732562/addon/SDCard/sdhost.cpp#L180-L193

The pin 34-39 are set to `GPIOModeInput` mode, which should instead be `GPIOModeAlternateFunction3`.

Setting the pin 34-39 to `GPIOModeInput` mode does not cause problem in Circle's WLAN sample app, because it will drive SD Host first to load the firmware. However, if the WLAN driver is started before the SD Host driver, the latter one will make the WLAN driver failed. In my case, we (my teammate and I) ported Circle to a microkernel OS, using the C++ classes to construct different drivers. We put the WiFi firmware into a ramdisk so we can run WLAN driver before SD driver, then the problem occurred.

### DIFF
--- a/addon/SDCard/sdhost.cpp
+++ b/addon/SDCard/sdhost.cpp
@@ -184,7 +184,7 @@ CSDHOSTDevice::CSDHOSTDevice (CInterruptSystem *pInterruptSystem, CTimer *pTimer
 	for (unsigned i = 0; i <= 5; i++)
 	{
 		m_GPIO34_39[i].AssignPin (34+i);
-		m_GPIO34_39[i].SetMode (GPIOModeInput, FALSE);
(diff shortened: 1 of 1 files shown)


rsta2:
I understand the problem with the SDHOST driver in your specific use case. Thanks for the correction!

richardchien:
Sorry I forgot to select the develop branch. I will remember next time! Thanks for the cherry-picking.

PR #259, opened 2021-10-30
Add support for Pi Zero 2 W Wi-Fi firmware

dwhinham:
This PR was originally going to include the machineinfo stuff aswell, but I saw you've already taken care of that! :slightly_smiling_face:

These changes get Wi-Fi working for me - everything seems to work fine; just a new chip revision and firmware blob.

The current WLAN firmware SHA in the Makefile seems new enough to contain the correct firmware.

Cheers!


rsta2:
Great that you found this out! Thank you.

Shouldn't we add the .clm_blob file for this firmware? I think, it contains WiFi regulatory data for specific countries. Maybe some WiFi channels are not usable in some countries otherwise?

dwhinham:
No problem! Apologies for missing the .clm_blob, you're absolutely right.

I've amended the commit and re-tested successfully.

rsta2:
Thanks Dale and no problem of course! It's great that WiFi is working now on this small powerful machine with Circle too.

PR #260, opened 2021-11-07
Enable PWM audio via GPIO on Zero 2 W

dwhinham:
These commits allow you to use `USE_PWM_AUDIO_ON_ZERO` and `USE_GPIO{18|19}_FOR_{LEFT|RIGHT}_PWM_ON_ZERO` to gain PWM audio output on the new Zero 2 W.

In `sysconfig.h`, I wrapped the PWM `#defines` in a `#if RASPPI == 1 || RASPPI == 3` block so that they are picked up when compiling with `RASPPI=3` as we do for Zero 2 W. This seemed to be the simplest way to achieve this, but if you'd prefer another way I'd be more than happy to respond to your feedback.

Cheers!

### DIFF
--- a/include/circle/sysconfig.h
+++ b/include/circle/sysconfig.h
@@ -83,7 +83,7 @@
 
 ///////////////////////////////////////////////////////////////////////
 //
-// Raspberry Pi 1 and Zero
+// Raspberry Pi 1, Zero (W) and Zero 2 W
 //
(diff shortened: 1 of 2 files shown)


rsta2:
Thanks for this contribution! I think, it's OK to add this to *sysconfig.h* in this way.

PR #312, opened 2022-08-31
Flashy HC-06 Bluetooth Support

elmfrain:
### Why I Made This?

I didn't have a USB to serial adapter to flash my software to the Raspberry Pi (but I did have a HC-06 Bluetooth module). I just set the serial port to a Bluetooth target (`rfcomm` on Linux, `COM` on Windows). However, it keeps on failing during transfer and it needed a fix.

### What About the HC-05?

The HC-05 works just fine with Flashy; I was able to upload my programs to the Pi successfully. Additionally, it works with a baud rate of 1382900 bits/s, so it is decently fast over air. Just make sure that the wires connecting the module to the PI isn't too long.

### What's Wrong With the HC-06?

By default, Flashy overloads the HC-06 when it sends mass amounts of data, and it causes a checksum or format error during transfer. Honestly, though, I don't exactly know how it fails. What it needs is data delivered in small pieces, or in this case, in records.

### The Fix

I propose this solution: whenever the user wants to use a HC-06 module as their flasher, they will set a `--hc06` flag. This will make Flashy flush each data record individually, for it prevents transfer failure this way. It may be slower, but it allows the module to decipher the stream of data more reliably. I was able to consistently upload my code to the PI multiple times without any failures.

**One thing to note though:** For both the HC-05 and HC-06, the Bluetooth driver does not block when fed data (at least on Linux). This means that Flashy sends the data almost instantaneously and then hangs, something like this:
```c++
Sending..........................ok
Finished in 0.1 seconds. // All that happens very quickly

Sending go command...  // But then appears to hangs at this point
```
This is normal. During the 'hanged' time it is when the data transfer is taking place, and it will finish up after awhile.

### DIFF
--- a/tools/flashy/flashy.js
+++ b/tools/flashy/flashy.js
@@ -20,6 +20,7 @@ let rebootMagic = null;
 let rebootDelay = null;
 let monitor = false;
 let noFast = false;
+let usingHC06 = false;
 let goDelay = 0;
 
 // The currently open serial port
@@ -208,7 +209,9 @@ function binary_encoder()
             if (inbyte == 0x3a)     // ":"
             {
                 // Flush buffer?
-                if (buffer_used > buffer.length - 1024)
+                // When using an HC-06 module as a serial device,
+                // flush on every record to fix transfer failure.
+                if (buffer_used > buffer.length - 1024 || usingHC06)
                     await flush();
 
                 // start of new record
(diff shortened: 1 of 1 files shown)


rsta2:
I added your commit on the *develop* branch. It will be part of the next release. Thanks for this contribution!

rsta2:
I added a link to this PR to the file *doc/bootloader.txt* on the *develop* branch to give users of these Bluetooth modules the necessary info. So this can be closed.

toptensoftware:
Hi @elmfrain,

I'm the author of flashy and just wanted to let you know I've just done a complete rewrite of it with lots of new features - Flashy v2.  In the process I removed your changes because I'm hoping they won't be needed...

Version 2 uses a much stronger packet encoding for all comms and after sending one packet, waits for ack before sending the next.  This should alleviate the issue you described above.

Since I don't have HC-06 here to test with I'd be interested to hear how it works for you.  If you get issues similar to before you can try using the `--packetSize:N` option to send in smaller packets.  The default is 4096.

See here for more https://github.com/toptensoftware/flashy

Brad

PR #325, opened 2022-11-02
Update diskio.cpp

bcs1234:
Implemented functionality needed by formatting Usb drives.

### DIFF
--- a/addon/fatfs/diskio.cpp
+++ b/addon/fatfs/diskio.cpp
@@ -16,6 +16,8 @@
 #include <circle/types.h>
 #include <assert.h>
 
+#include <circle/usb/usbmassdevice.h>
+
 #if FF_MIN_SS != FF_MAX_SS
 	#error FF_MIN_SS != FF_MAX_SS is not supported!
 #endif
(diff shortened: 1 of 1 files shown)


rsta2:
Thanks for this PR. Unfortunately with this modification each use of the FatFs library would require the whole USB code to be included in a binary, even when only a SD card is used instead of USB. There are also sample programs in Circle, which use FatFs and currently do not specify the USB library in the LIBS variable in the Makefile. They wouldn't build any more without modification. That's why I cannot accept this PR immediately. I have to think about, if there is a solution.

bcs1234:
Thanks for the quick response. I didn't think of that problem.
I think the cleanest solution for this is to declare a macro in ffconf.h, with which this functionality is turned off by default.

To use disk formatting, it is necessary to turn on the FF_USE_MKFS macro (FF_MULTI_PARTITION is recommended) and then there could also be a FS_USE_USB_LIB macro.

Have a nice day.

rsta2:
I've found another solution, which is more generic, thought. There is a new virtual method `CDevice::GetDevice()`, which returns the total size of a block device in bytes, and which is implemented in the class `CUSBBulkOnlyMassStorageDevice`. So the USB library is only needed, if USB operation is used. This saves a lot of space for SD card only applications. The option `FF_USE_MKFS` must be still enabled to use `f_mkfs()`, because that saves another 4K, not much but enough for a rarely used function.

Thanks again for this contribution! I merged it to the *develop* branch and also applied two recent patches to FatFs. This is the first time, I have created a file system on an USB flash drive using Circle. ;)

PR #331, opened 2022-12-04
Fix missing field initializers in soundcontroller.h

dwhinham:
This is a trivial commit which silences a compiler warning from GCC (`-Wmissing-field-initializers`).

Thanks!


rsta2:
Ah, forgot this. Thank you!

PR #347, opened 2023-01-24
Add ctrl arrows

kevinboone:
This patch adds support for ctrl+arrow, ctrl+page, ctrl+home and ctrl+end. I've expanded the keymap table by an additional row, so it should now be easy to add other ctrl+key combinations if required.

This change will expand binary size of a kernel by a couple of kilobytes. Because the changes required are distributed across a number of source files, I can't see any easy way to make the added ctrl handling configurable. However, for may application these ctrl keys are essential, so I would be willing to live with the small expansion.

The new key combinations generate the same codes as a USB keyboard on Linux. So, for example, ctrl+right is "esc [ 1 ; 5 C". 

To be frank, I've only tested the DE and US layouts but, so far as I can see, the changes are the same for all layouts.


rsta2:
Thank you for this contribution! I merged it manually to the *develop* branch, where development takes place.

PR #352, opened 2023-02-15
Make KERNEL_STACK_SIZE configurable

mjrgh:
This is a proposal to make KERNEL_STACK_SIZE configurable via sysconfig.h or Config.mk, in parallel to KERNEL_MAX_SIZE.  It removes

#define KERNEL_STACK_SIZE 0x20000

from memorymap.h and memorymap64.h, and moves it instead to sysconfig.h, made conditional on the symbol not already being defined (so that it can be overridden).  I ran into this because I need a larger stack size than the default, and it's of course better not to have to directly edit the base code in my local repos clone.

### DIFF
--- a/include/circle/memorymap.h
+++ b/include/circle/memorymap.h
@@ -39,7 +39,6 @@
 
 #define PAGE_SIZE		4096				// page size used by us
 
-#define KERNEL_STACK_SIZE	0x20000				// all sizes must be a multiple of 16K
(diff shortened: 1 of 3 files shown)


rsta2:
Thanks for this PR. I merged it manually, because I did not want to have the merge commits from the *master* branch on *develop* again. There was a typo in the `KERNEL_STACK_SIZE` macro name in *sysconfig.h*. I fixed it.

PR #353, opened 2023-02-15
Add wait loop to serial port write

mjrgh:
I've found that CSerial can be problematic for debug logging, because it's easy to write data faster than the underlying port can accept it.  Overruns are currently summarily discarded.

This change adds a busy-wait to CSerial::Write(uint8) when an interrupt handler is present.  When there's an interrupt handler, bytes in the buffer should continue to be sent asynchronously while the foreground thread is looping, so space in the buffer should eventually free up.  The loop uses a CTimer::SimpleusDelay between attempts, and limits the total wait time (to an arbitrary maximum number of retries) so that it doesn't lock up if the UART isn't actually sending anything.

I'm not sure if the existing discard-on-overrun behavior was intentional so as to make the routine explicitly non-blocking.  I don't see anything to suggest that callers should expect it to be, but if that was the intention, this change wouldn't be desirable.  In that case, though, maybe you could take this code and move it to a new "blocking write" method, alongside the existing one, so that callers could choose which one to use according to their needs.

### DIFF
--- a/lib/serial.cpp
+++ b/lib/serial.cpp
@@ -18,6 +18,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 //
 #include <circle/serial.h>
+#include <circle/timer.h>
 #include <circle/devicenameservice.h>
 #include <circle/bcm2835.h>
 #include <circle/memio.h>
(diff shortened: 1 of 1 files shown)


rsta2:
For debugging output normally the blocking variant of `CSerialDevice` (without interrupt) is used. As you wrote, with your modification the `Write()` method may block, which is not intended with the interrupt variant. Instead this method returns the number of characters, which were successfully sent, so the caller can handle the case, that some chars were not sent. One could also increase the [serial buffer size](https://github.com/rsta2/circle/blob/master/include/circle/serial.h#L54), if it is required to use the interrupt variant. So thanks, but I cannot accept this PR.

PR #354, opened 2023-02-16
DMA4 fixes

mjrgh:
This has some changes that I've found necessary for stable memory-to-memory transfers with DMA4 on Pi 4B.  The current Circle dma4channel.cpp is very unstable for me - it seems to have a probability of about 0.1% of crashing the Pi on any given transfer.  That doesn't sound like it would happen a lot, but I'm using this for video updates, so it gets exercised heavily in my setup.  The crash typically manifests in 30-60 seconds of up time.

I had a previously working setup with the same client code, based on my own DMA4 implementation that I wrote before Circle had DMA4 support.  My version was working reliably, so I figured I'd go back and compare our respective implementations to see if I could track down the problem.   I think I've found a minimal set of changes to Circle's version that gets it working reliably for me.  That's what this PR represents.

Explanation of the changes:

- Removed the TI4_WAIT_RD_RESP | TI4_WAIT_RESP flags from the transfer setup for memory-to-memory copies.  The BCM2711 documentation on these is vague, but it seems to talk about them mostly in the context of peripheral device FIFOs, so maybe they're not supported for memory-to-memory transfers.  Experimentally, TI4_WAIT_RD_RESP seems to be the real culprit; leaving TI4_WAIT_RESP in place seems okay.  But it also doesn't seem to be necessary, given that my old implementation used 0 for the transfer flags and worked reliably for me.  I tried measuring the timing with and without TI4_WAIT_RESP, and I don't see any difference.  I'd think that any effect it was having on bus timing or loading would show up in the transfer timing.  Maybe it's a no-op for mem-to-mem.

- Added PeripheralEntry() / PeripheralExit() bracketing wherever DMA controller registers are updated (ctor, Start(), and Wait()).  I'm not sure if these are strictly necessary, but given the random nature of the crashes, it seems safer to include them, and I note that the regular DMA code in dmachannel.cpp uses them consistently.

- Note that there are a bunch of other methods that probably should have the PeripheralEntry/Exit bracketing as well.  I didn't want to change any code that I'm not actually exercising, though, so I left the others as-is.

- In the ctor, explicitly cleared the channel_CB (control block) register before enabling the channel, just in case there's previous garbage in the register.  (I don't think this had anything to do with my crashes, since it only affects initialization, but it seems prudent. The BCM2711 documentation specifically mentions that the DMA controller will blindly process whatever's in the control block when enabled, even if it's random garbage.)

- Changed the control block alignment to 32 bytes. This is just a minor thing unrelated to the instability, but the existing version does 256-BYTE alignment of the control block, whereas the BCM2711 documentation only calls for 256-BIT alignment (== 32 bytes).  I noticed that in comparing our versions,  It doesn't do any harm to align to the larger multiple of 32, but it does waste a little RAM.

### DIFF
--- a/lib/dma4channel.cpp
+++ b/lib/dma4channel.cpp
@@ -108,22 +108,27 @@ CDMA4Channel::CDMA4Channel (unsigned nChannel, CInterruptSystem *pInterruptSyste
 	m_pCompletionParam (0),
 	m_bStatus (FALSE)
 {
+	PeripheralEntry();
+
 	assert (m_nChannel >= DMA4_CHANNEL_MIN);
 	assert (m_nChannel <= DMA4_CHANNEL_MAX);
 	assert (   (read32 (ARM_DMA4CHAN_DEBUG (m_nChannel)) & DEBUG4_VERSION_MASK)
 		>> DEBUG4_VERSION_SHIFT == DMA4_VERSION);
 
-	m_pControlBlockBuffer = new (HEAP_ANY) u8[sizeof (TDMA4ControlBlock) + 255];
+	m_pControlBlockBuffer = new (HEAP_ANY) u8[sizeof (TDMA4ControlBlock) + 31];
 	assert (m_pControlBlockBuffer != 0);
 
-	m_pControlBlock = (TDMA4ControlBlock *) (((uintptr) m_pControlBlockBuffer + 255) & ~255);
+	m_pControlBlock = (TDMA4ControlBlock *) (((uintptr) m_pControlBlockBuffer + 31) & ~31);
 	m_pControlBlock->nReserved = 0;
 
+	write32 (ARM_DMA4CHAN_CONBLK_AD (m_nChannel), 0);
 	write32 (ARM_DMA_ENABLE, read32 (ARM_DMA_ENABLE) | (1 << m_nChannel));
 	CTimer::SimpleusDelay (1000);
 
 	write32 (ARM_DMA4CHAN_DEBUG (m_nChannel), DEBUG4_RESET);
 	CTimer::SimpleusDelay (1000);
+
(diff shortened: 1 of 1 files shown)


mjrgh:
Just a little by-the-way on how great the DMA4 peripheral is.  In my testing, it's about 10X faster than the legacy DMA channels for my video buffer transfers.  The fastest way I've found to do video updates on the Pi4 is to render into a regular RAM buffer, and then do a DMA4 transfer into the frame buffer.  This is much faster than writing directly into the frame buffer from CPU code - apparently the DMA4 controller has faster access to the VideoCore memory than the CPU does.  This was counterintuitive to me; it seemed like the most efficient thing would be to set up a double frame buffer, do CPU rendering directly into the off-screen half of the frame buffer, and swap the frame buffer frames at vsync.  It's actually much faster to allocate a regular heap buffer, render into that, DMA4 transfer from the heap buffer to the off-screen half of the frame buffer, and then do the swap at vsync.  You'd think the extra copy would double the time, but it's actually several times faster.  Direct CPU access to the frame buffer must trigger a lot of wait cycles.

rsta2:
Thank you, also for the detailed info! Please note, that the `Peripheral{Entry|Exit}()` macros are [NOP's](https://github.com/rsta2/circle/blob/master/include/circle/synchronize.h#L130) on the Raspberry Pi 2-4.

mjrgh:
> Please note, that the `Peripheral{Entry|Exit}()` macros are [NOP's](https://github.com/rsta2/circle/blob/master/include/circle/synchronize.h#L130) on the Raspberry Pi 2-4.

Ah, I didn't realize that.  My problem must have boiled down to that wait flag in the transfer information, then.  Thanks for accepting the PR!

PR #356, opened 2023-03-05
Rules.mk: Add option to gzip-compress the kernel

dwhinham:
This is quite a trivial commit but may be useful to those looking to reduce the final kernel image size.

Cheers :slightly_smiling_face: 

---

The Raspberry Pi bootloader is capable of booting kernels compressed with gzip. This can help improve startup times with larger kernels containing static data.

Allow the user to define GZIP_KERNEL = 1 so that the final kernel image is replaced with a gzip-compressed file.


rsta2:
Thank you!

PR #362, opened 2023-04-25
DrawImageRect uses display stride instead of image stride

AterialDawn:
The current behavior copies pixels from the source image using the display stride, not the image stride. This causes the RPi to draw garbage on screen if the source image is not the same resolution as what is displayed on screen.

This is what is currently drawn on screen now
Untitled-1

This is what is expected to be drawn
Untitled-2


rsta2:
Oops. Thank you for the fix! I merged it manually to the *develop* branch, where development takes place. I can imagine, that this was not easy to find.

PR #370, opened 2023-05-28
Added Synchronization Mutex

toptensoftware:
CSynchronizationMutex - a mutex class for scheduled tasks with re-entrancy.


toptensoftware:
Ignore this, didn't notice already merged "CMutex" class.

PR #373, opened 2023-05-28
Added CDevice::IOCtl method

toptensoftware:
Added a IOCtl method to the CDevice class.  This completes the typical set of methods that a posix like device would normally support. 

While not actually used by circle , I'm using this on a "RamDisk" device in my project and map FatFS disk_ioctl calls to it.  

(I understand if you don't want to include this in circle, but it's just one virtual no-op method and by having it included in the base CDevice class it saves me having to implement my own separate and conflicting CDevice class. ).

### DIFF
--- a/include/circle/device.h
+++ b/include/circle/device.h
@@ -52,6 +52,13 @@ public:
 	/// \note Supported by block devices only
 	virtual u64 GetSize (void) const;
 
+    /// \param cmd The IOCtl command to invoke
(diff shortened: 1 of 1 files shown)


rsta2:
Implemented on the *develop* branch slightly modified. Thanks for suggesting this, I thought about adding an `IOCtl()` method earlier, but had no concrete application for it so far.

PR #374, opened 2023-05-28
Fix for flashy serial port I/O errors under WSL

toptensoftware:
Fixes an issue with "flashy" where in some circumstances serial port I/O errors occur if flashy is launched directly from a WSL session.

Not sure when this started happening, nor why, but might be related to Windows 11 or possibly a newer version of WSL.

Since the problem only seems to happen in a process launched directly from WSL, the fix is simple, albeit hacky - the script simply respawns itself to run in a new child process.


rsta2:
Manually merged to the *develop* branch. Thanks for the fix!

PR #375, opened 2023-05-29
Added SysConfig.h variable for name of "main()"

toptensoftware:
This is a simple change to allow redefining the name of the "main()" proc.

Use case: I have an intermediate library that wraps circle and has its own startup/shutdown code.  This change enables circle to call my library's entry point (eg: -DMAINPROC=mylib_main) to do its own initialization before on-calling the application's "main()".

(I could require apps that use my library to have a different main proc name, but I like the idea of keeping this transparent and leaving it as "main()");

### DIFF
--- a/include/circle/sysconfig.h
+++ b/include/circle/sysconfig.h
@@ -417,6 +417,19 @@
 
 #endif
 
+
+// Sets the name of the "main()" entry point function that will be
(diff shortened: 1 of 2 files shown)


rsta2:
Thanks!

PR #381, opened 2023-06-10
Blackaddr/add tftp kernel helpers

Blackaddr:
For my application, I need to be able to detect when a TFTP transfer is in progress, and if a "kernel*" file is received (i.e. a kernel image) as a special case since normally this means to schedule a reboot.

This PR adds the ability to query CTFTPDaemon if an xfer is in progress, and I added a sticky flag to indicate if a kernel image has been received. A private flag is first set when the file xfer is complete, and the public flag (via API function) is set when the TFTP connection has been properly closed.


Blackaddr:
Let me know if the target for the PR should be 'master' or 'develop'.

rsta2:
Thanks. Isn't it possible to implement this application-specific functions by deriving a user class from `CTFTPFatFsFileServer` or `CTFTPFileServer` in *addon/tftpfileserver* without much effort. I think, these functions (especially the detection of kernel images) do not fit well into the generic `CTFTPDaemon` class. BTW. It's not necessary to use `Atomic*()` here, because the scheduler is cooperative and interrupts or secondary CPU cores are not involved.

rsta2:
I've added a method `CTFTPDaemon::UpdateStatus()`. By overriding this method one can follow the progress of a TFTP transfer and derive the needed information. For instance, when `UpdateStatus(StatusWriteCompleted, "kernel.img")` is called, you will know, that a kernel image has been successfully received. I think this is a more clean solution. In the hope that it helps you.

dlbogdan:
Fully agree! That is a nice and elegant solution.

Blackaddr:
@rsta2 In order to be useful in an application, UpdateStatus() would need to communicate outside the scope of CTFTPDaemon. That requires creating a class to inherit CTFTPFileServer in my case, and adding appropriate points to other data, etc.

Would you be willing to add a default implementation to the UpdateStatus() that stores the status and filename value in CTFTPFileServer members, and either make them public, or provide getter functions? This seems more elegant than all the overloading and class inheritance. With this, I can simply query the last status directly with no need for the overriding, inheritance, etc.

rsta2:
@Blackaddr I cannot see a problem in inheriting a user specific TFTP file server class from `CTFTP[FatFs]FileServer`. This can be quickly done and added to your application. Unfortunately Circle cannot fulfill all user requirements in its classes, it can only try to provide a way to implement user specific features, which has been done here.

rsta2:
Attached you find a modified *addon/tftpfileserver/sample*, which contains a class `CMyTFTPFileServer`, which implement your extended API, based on the `UpdateStatus()` method. I modified `UpdateStatus()` once again to be not `const` any more. I have to admit, that without this the implementation would have been more difficult.

[mytftpfileserver.zip](https://github.com/rsta2/circle/files/11800412/mytftpfileserver.zip)

PR #391, opened 2023-07-11
Update windows-build.txt ARM toolchain link

ericfont:
link changed to much shorter.  I think don't bother with specific filename cause that's a moving target.  Instead I think it is more useful to write the AArch64 triplet in addition to the AArch32 triplet.

Note: I still did have to go to a subdirectory https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads/12-2-rel1 to get an AArch64 target, but I tested building the circle example web server with it and it works.

### DIFF
--- a/doc/windows-build.txt
+++ b/doc/windows-build.txt
@@ -17,9 +17,10 @@ First you have to download and install the following software:
 * GNU make for Windows (Complete package, except sources, Setup)
   from: http://gnuwin32.sourceforge.net/packages/make.htm
 * Toolchain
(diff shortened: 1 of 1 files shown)


rsta2:
Yes, Rel12.2.1 is already the recommended toolchain for Circle builds, but had not been updated in *doc/windows-build.txt* yet. Thanks for that! Your commit is on the *develop* branch. I had to apply it manually, because commits to *master* are only done for an release (see *CONTRIBUTING.md*).

PR #392, opened 2023-07-12
Update bootloader.txt to make bootloader64 if building for AArch64

ericfont:
need to additionally run "make bootloader64" if want to run on AArch64 rpi4, because the "make all" command in the Makefile is only:

"all: firmware bootloader"

(Of course alternatively the Makefile could be changed to automatically detect whether "AARCH" from Config.mk is "32" or "64".)

### DIFF
--- a/doc/bootloader.txt
+++ b/doc/bootloader.txt
@@ -30,13 +30,17 @@ using the serial interface.
 4. The Circle project libraries used in your application have to be build
 manually using "./makeall" as described in the main README.md file.
 
-5. You have to prepare a SD card, which starts your Raspberry Pi in bootloader
-mode. Go to the boot/ subdirectory and enter:
+5. You have to prepare a SD card with a FAT partition, which starts your
+Raspberry Pi in bootloader mode. Go to the boot/ subdirectory and enter:
 
 	make all
 
-After completion copy the files from boot/ to the SD card (copy the file
-"config.txt" only for AArch64 operation), which must have a FAT partition.
+5b. If building for AArch64, then additionally enter:
+
+	make bootloader64
+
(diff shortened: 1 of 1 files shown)


ArnoldUK:
> (I'm embarrassed to admit how many hours of time I wasted trying to figure out why bootloader wouldn't work when building for AArch64, until I finally looked at the Makefile's all command!)

There is no need to edit the main `Makefile` to change the build target. Simply edit the `Config.mk` and change the following lines:
```
AARCH = 64
RASPPI = 3
PREFIX64 = aarch64-none-elf-
```

All explained in the readme docs.

ericfont:
Just to reiterate, the problem is bootloader.txt says:

https://github.com/rsta2/circle/blob/2bdfcf27bd81535205396a1e9e08e80108a0a527/doc/bootloader.txt#L33-L39

But with a 64-bit `Config.mk`, following those instructions fails to produce the important files `kernel8-rpi4.img` and `kernel8.img` which are referenced by `config64.txt`, so when the user copies the /boot folder, those necessary `.img` files aren't copied, so the bootloader won't work.

ericfont:
the relevant lines in the Makefile are:

https://github.com/rsta2/circle/blob/2bdfcf27bd81535205396a1e9e08e80108a0a527/boot/Makefile#L33-L45

and hence the `make all` command only runs the 32-bit `bootloader:` code and not the `bootloader64:` code.

rsta2:
Thanks for the update. The problem is, that some users only have a 32-bit toolchain installed and some have both. But maybe the SD card creation process needs a closer look.

rsta2:
I've added targets for the creation of SD cards for 32- / 64-bit operation or both to *boot/Makefile*. This is described in *boot/README* (last section).

PR #411, opened 2023-11-13
Add CTimer::GetClockTicks64 function

### LINKED ISSUE #410: GetClockTicks without wrap
The current implementation of `CTimer::GetClockTicks` may wrap - as documented.
However - at least when using the physical counter - this could be simply avoided by also using the value of `ARM_SYSTIMER_CHI` which already exists `bcm2835.h`.
When using the physical counter, a simple implementation (inspired by [LLD's kernel](https://github.com/rockytriton/LLD/blob/main/rpi_bm/part17/src/kernel/timer.c#L34-L45)) could look like this:

```cpp
u64 CKernel::GetClockTicksHiLo() {
  PeripheralEntry();

  u32 hi = read32(ARM_SYSTIMER_CHI);
  u32 lo = read32(ARM_SYSTIMER_CLO);

  // double check hi value didn't change when retrieving lo...
  if (hi != read32(ARM_SYSTIMER_CHI)) {
    hi = read32(ARM_SYSTIMER_CHI);
    lo = read32(ARM_SYSTIMER_CLO);
  }

  PeripheralExit();

  return static_cast<u64>(hi) << 32 | lo;
}
```

I am just pointing this out because I was wondering why something like this is not offered yet by circle.
It is probably because of a missing counterpart when not using the physical counter (?)
If you want me to create a PR, I can gladly do so.

ThexXTURBOXx:
This should support all options the same way `CTimer::GetClockTicks` does.
The changes are minimal, except for the physical counter part.

I hope that is okay :)
Also, thanks for the quick reply!

Closes #410

### DIFF
--- a/include/circle/timer.h
+++ b/include/circle/timer.h
@@ -66,6 +66,9 @@ public:
 
 	/// \return Current clock ticks of an 1 MHz counter, may wrap
 	static unsigned GetClockTicks (void);
+
+	/// \return Current clock ticks of an 1 MHz counter (continuous)
+	static u64 GetClockTicks64 (void);
 #define CLOCKHZ	1000000
 
 	/// \return 1/HZ seconds since system boot, may wrap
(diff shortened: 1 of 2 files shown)


rsta2:
Thank you.

PR #443, opened 2024-03-08
Some more test to GPIO timing... still WIP

pottendo:
Hi @sta2, 
I hope I did everthing right with the fork & branch and pull request.
I've reworked and added more tests to the GPIO timing.
Couldn't test Pi4, as the tree currently fails to build for Pi4 due to some missing define in I2C.
Further, I don't know whether tests make sense and can work on all PI models like this (masks, etc.).
bye, pottendo


rsta2:
Thanks, but it doesn't work this way. I had again apply your modifications manually, because I do not want to have about 30 doubled commits in the repo. Do not apply commits from the Circle repo after creating the branch, you are working on, to this branch. And please, when you create the PR, select the *develop* branch, not *master*.

Also you have to accept, that your method to handle the screen/serial switching is not the way, this is currently handled in Circle, where there is a `SCREEN_HEADLESS` option and the `logdev=` cmdline option.

pottendo:
Sorry, for the inconvenience. I obviously missed something when reading the git workflow.

I now use logdev= in cmdline.txt, which works. 
The problem in my setup that pi5 fails with the screen and then no logging appeared on the serial. Even with logdev set...
I didn't know about SCREEN_HEADLESS, need to check.
Br pottendo

rsta2:
SCREEN_HEADLESS is exactly for that purpose, when there is no screen attached 
to the RPi 4 or 5. You can define it with -d, when calling "configure". Of 
course you can use a different method for your own projects, but for code, 
which should go into Circle, this method should be used.

PR #450, opened 2024-05-05
Add ST7789 character device overlay

diyelectromusic:
This is a relatively shallow overlay layer using the existing addon/display/st7789 graphics driver to provide the same character/console interface as the ssd1306 and hd44780 displays.

To achieve this it has had to add an option for rotation and setting the width of characters in the main ST7789 display code too. 
 There is also a new sample application showing how it can be used.

If this is not wanted within circle itself, then I'll add it into MiniDexed directly.  Ultimately we want to support ST7789 displays in MiniDexed and this seemed like the simplest way to do it.

Note: I've set the sample application to use the same hardware configuration as the original ST7789 sample application, but this is untested as it isn't the hardware I have.

I've tested it with a Pimoroni Pirate Audio add-on with built-in ST7789 display.

Many Thanks,
Kevin

### DIFF
--- a/addon/display/Makefile
+++ b/addon/display/Makefile
@@ -4,7 +4,7 @@
 
 CIRCLEHOME = ../..
 
-OBJS	= hd44780device.o st7789display.o chardevice.o ssd1306device.o
+OBJS	= hd44780device.o st7789display.o chardevice.o ssd1306device.o st7789device.o
 
 libdisplay.a: $(OBJS)
 	@echo "  AR    $@"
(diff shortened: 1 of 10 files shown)


rsta2:
Many thanks for this contribution, Kevin! Of course, it's good to have this in Circle.

rsta2:
Forgot to mention: I tested both samples successfully with my no-name 240x240 
pixels ST7789 display.

diyelectromusic:
I'm glad you did that as later in the day my display developed a hardware fault and I'm pretty sure it wasn't anything I did!!  But it is great to know it worked for someone else :)

(I think it is a dodgy connection to part of the display - I'm getting hardware fault lines)

Thanks - I can add that to MiniDexed now .

Kevin

PR #451, opened 2024-05-06
Fix for character handling rotationfor asymetric displays

diyelectromusic:
My apologies Rene!  There was a bug in the character handling for rotations which I'd missed as my displays were all square!  Hopefully this sorts that out.

Sorry about that! I think that is ok now... but I don't have a wide range of displays to test it on unfortunately.  Hopefully once in the branch a few others might be able to give it a go.

Many Thanks,
Kevin

### DIFF
--- a/addon/display/sample/st7789char/kernel.cpp
+++ b/addon/display/sample/st7789char/kernel.cpp
@@ -40,7 +40,8 @@
 #define COLS  15	// Max 40
 #define ROWS  4		// Max 4
 #define ROT   0 	// 0,90,180,270
-#define WIDECHARS	true	// Set to false for thin characters
+#define WIDECHARS	TRUE	// Set to false for thin characters
+#define TALLCHARS	TRUE	// Set to false for short characters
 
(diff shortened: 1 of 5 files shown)


rsta2:
No problem, Kevin, that happens. It's good, that you have found the problem. 
I will test the update tomorrow and will merge it then.

Rene

diyelectromusic:
Thinking about this a bit more and seeing as there are such a range of possible display sizes, I'm going to add one more configuration parameter, so hold off a while and I'll get that update to you later today.

Part of the issue is not knowing what screens/configurations were used with the original code to know why, for example, it was using double height/double width text in the first place!  With smaller displays that doesn't look like it will be needed.

Thanks,
Kevin

rsta2:
I think, that's good. I merged it. I tested both samples successfully (character sample with 6 rows and rotation 90). I added two small other changes:

* Allow up to 10 character rows in chardevice.h
* Clear display after switching it on in `CST7789Display::Initialize()`. Otherwise the clear was not reliable here.

Thanks again,
Rene

PR #470, opened 2024-09-08
Fixed LVGL Makefile.

clickworkorange:
LVGL would not build as pulled; the reference to `lvgl.h` was nested one level too deep:

```
make: *** No rule to make target 'lvgl/lvgl/lvgl.h', needed by 'lvgl.o'.  Stop.
```

I fixed that, tidied up a bit, and added a check to warn if the LVGL submodule is missing.

### DIFF
--- a/addon/lvgl/Makefile
+++ b/addon/lvgl/Makefile
@@ -3,11 +3,13 @@
 #
 
 CIRCLEHOME = ../..
+LVGL_DIR = lvgl
 
-LVGL_DIR = .
-LVGL_DIR_NAME = lvgl
+ifeq (,$(wildcard $(LVGL_DIR)/lvgl.mk))
+    $(error LVGL submodule not checked out!)
+endif
 
-include $(LVGL_DIR)/lvgl/lvgl.mk
+include $(LVGL_DIR)/lvgl.mk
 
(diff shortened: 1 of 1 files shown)


rsta2:
Thanks. But the file [lvgl/lvgl.mk](https://github.com/lvgl/lvgl/blob/v8.3.11/lvgl.mk) references the make variable `LVGL_DIR_NAME`, which is not set in your modified Makefile.

clickworkorange:
So it is, my mistake, sorry! Though it was removed on [22 July 2022](https://github.com/lvgl/lvgl/commit/e42f01a38ad43140cdeb99799c737e9f5e668103#diff-48f30bcb63b179deef2af4c97c6ed558dcf803637390d026acfa4f75782f6242), and replaced with 
```
LVGL_PATH ?= ${shell pwd}/lvgl

CSRCS += $(shell find $(LVGL_PATH)/src -type f -name '*.c')
...
```
Strange that change wasn't included in 8.3.11. I didn't catch it because `LVGL_DIR_NAME` evaluates to an empty string and the addon builds without error with my modified Makefile. Never mind, apologies for wasting your time with this!

rsta2:
No problem.

PR #472, opened 2024-09-13
Fix backlight issue for older RPi 7" Touchscreen Display

SteveSpencer:
Older RPi 7" Touchscreen Display controller board (revision 1.0) does not seem to correctly set the backlight when starting.
This change adds 'backlight=' as a possible entry in cmdline.txt. If a value between 1 and 255 (inclusive) is supplied, this value is used to set the backlight when the BCM framebuffer is initialised.


rsta2:
That's a clean solution. I merged it. Thank you!

PR #480, opened 2024-09-26
Cstring extension

SteveSpencer:
Added += operators to CString
Added c_str () and Append (const char chChar) members to CString


rsta2:
I added your commits manually, because of the additional commits, which I don't wanted to have on the *develop* branch. Thank you for this contribution!

PR #528, opened 2024-12-26
Small code comment correction (ARM_CONTROL register bit numbers)

petemoore:
This is just a minor doc fix. I've created an upstream PR too: https://github.com/raspberrypi/tools/pull/134

See https://datasheets.raspberrypi.com/bcm2711/bcm2711-peripherals.pdf page 93

The assumption here is that the PDF is correct and the code comments are wrong, but I guess it could be the other way around...

### DIFF
--- a/boot/armstub/armstub8.S
+++ b/boot/armstub/armstub8.S
@@ -74,8 +74,8 @@
 _start:
 	/*
 	 * LOCAL_CONTROL:
-	 * Bit 9 clear: Increment by 1 (vs. 2).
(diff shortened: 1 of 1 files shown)


rsta2:
I think we should wait, until the PR has been handled in the raspberrypi/tools project. They should now it best.

rsta2:
There is no further reaction in raspberrypi/tools and I want to complete this now. I think we can trust the documentation, so I merged your correction. Thank you!

PR #540, opened 2025-01-26
addon/fatfs: Update to FatFs R0.15a + patch1

ThexXTURBOXx:
Latest patch from here: http://elm-chan.org/fsw/ff/patches.html


rsta2:
Thank you!

PR #541, opened 2025-02-02
Add support for SSD1305

probonopd:
Add support for SSD1305 by @Sterr1.

For details please see https://github.com/probonopd/MiniDexed/issues/808#issuecomment-2629132331

### DIFF
--- a/addon/display/ssd1306device.cpp
+++ b/addon/display/ssd1306device.cpp
@@ -134,11 +134,12 @@ boolean CSSD1306Device::Initialize (void)
 	assert(m_pI2CMaster != nullptr);
 
 	// Validate dimensions - only 128x32 and 128x64 supported for now
-	if (!(m_nHeight == 32 || m_nHeight == 64) || m_nWidth != 128)
+	if (!(m_nHeight == 32 || m_nHeight == 64) || !(m_nWidth == 128 || m_nWidth == 132))
 		return false;
 
+	const bool bIsSSD1305 = m_nWidth == 132;
 	const u8 nMultiplexRatio  = m_nHeight - 1;
-	const u8 nCOMPins         = m_nHeight == 32 ? 0x02 : 0x12;
+	const u8 nCOMPins         = (m_nHeight == 32 && !bIsSSD1305) ? 0x02 : 0x12;
 	const u8 nColumnAddrRange = m_nWidth - 1;
 	const u8 nPageAddrRange   = m_nHeight / 8 - 1;
 	// https://www.buydisplay.com/download/ic/SSD1312_Datasheet.pdf Pg. 51 Section 2.1.19


rsta2:
Thank you!

diyelectromusic:
This appears to be suggesting that if a user sets of width of 132 pixels rather than 128 then we make the leap that it is using a different driver chip - is that correct?

That doesn't seem like a very robust way of handling that...

Should we not perhaps be presenting alternative constructors for ssd1306-like displays that require different configurations?  Or at the very least a set of options in a parameter related to the display itself?

Are we confident these alternative chips have no other command registers that are different?

Aside: I've found that ssd1315 appears to "just work" with all ssd1306 parameters; but personally I'd like to find a way to support SH1106 displays too, but have only just started to take a look.

So basically, I'm interested in a more robust multi-driver option so I can potentially expand on it :)

Kevin

Sterr1:
The SSD1305 isn’t too different from the SSD1306, so the code changes are minimal. Who knows why the manufacturer suddenly made the buffer 132 pixels long while the image is 128. We’ll leave that on the manufacturer’s conscience. I encountered a situation where MT32-pi works in this configuration, but MiniDexed does not. So, I started looking for differences. The difference was only in 3 lines of code, which I added. Dale Whinham apparently used the original code from rsta2 and adapted it for SSD1305 as he saw fit, which I also added to the Circle code, as everything started working as needed. If you’re looking for ways to add SH1106, Dale Whinham has that driver in mt32-pi. You can also add it to Circle and test it. I worked with what I had. And if Dale Whinham managed to add another device by changing just a few lines of code, then that’s the simplest path. Of course, you could formalize this separately and write a separate driver for SSD1305, but why complicate things? It works, after all.

Dale Whinham wrote a separate driver for SH1106, which means the differences are significant. While studying the compatibility of SSD13xx displays, I found information that SH11xx displays are heavily simplified compared to SSD13xx and lack more than half of the features of SSD13xx.

Gene J.B.

Sterr1:
Well, add another variable bSSD1305, which will be 1 if 1305 is used and 0 if 1306 is used, and modify the code something like this:

```
if (bSSD1305 && (m_nWidth == 128)) {
    m_nWidth = 132
} 

```
But this will only complicate the code, even though it will be more correct. 
But existing applications would need to be significantly reworked to introduce another variable, add it to configuration files, and document all of this. Instead of simply writing: "If you have an SSD1305, use the SSD1306 settings but specify a width of 132 pixels." Which approach do you prefer?

rsta2:
I'm up to preparing a new release and cannot accept more changes at the 
moment.

Please note that the following classes will be deprecated in some future 
release, because there is a unified display interface now, which allows to use 
the CTerminalDevice class on any display driver, which inherits from CDisplay:

* CSSD1306Device
* CST7789Device
* CCharDevice

I accepted this PR, because there must be another modification for the class 
CSSD1306Display anyway to support the SSD1305. Please see the following 
article for more details:

https://github.com/rsta2/circle/discussions/380#discussioncomment-11303686

I will come back on it after the release.

rsta2:
[Adafruit wrote](https://forums.adafruit.com/viewtopic.php?p=149472#p149472) that the SSD1305 is completely different from the SSD1306. I guess with the code in this PR it works with the class `CSSD1306Device` in Circle, but it's not clear, if the new class `CSSD1306Display`, which provides the new `CDisplay` interface can be modified in the same way to support the SSD1305. `CSSD1306Display` will be the only supported display driver for the SSD1306 in the future.

So if someone, who owns a SSD1305-based display, wants to modify the class `CSSD1306Display` to add SSD1305 support, he/she has to test carefully, if it works with *sample/08-usbkeyboard*, *sample/41-screenanimations* and *addon/lvgl/sample* with both displays. I only have a SSD1306-based display and cannot do this.

I think, if it is possible to support both SSD1305 and SSD1306 in one driver, there should be another parameter for the constructor, which select the actual display controller type. An enum type could be defined in the class for this purpose.

Sterr1:
Displays not different. KIMI analyse datasheets and find minor differences.

_Commands unique to SSD1305:

Set Look Up Table (91h): Sets the color lookup table (LUT) for different colors and their brightness.
Set Bank Color of BANK1 to BANK16 (92h) and Set Bank Color of BANK17 to BANK32 (93h): Set colors for different banks (BANK1-BANK32).
Set Dim Mode setting (ABh): Controls brightness and contrast in dim mode.
Set Area Color Mode ON/OFF & Low Power Display Mode (D8h): Enables/disables area color mode and low power display mode.

Commands unique to SSD1306:

Set Fade Out and Blinking (23h): Manages fade-out and blinking modes.
Set Zoom In (D6h): Enables zoom-in mode for image enlargement.
Commands present in both devices but with slight differences:
Set Contrast Control for BANK0 (81h): SSD1305 supports contrast adjustment for BANK0, while SSD1306 is limited to monochrome display.
Set Display ON/OFF (AEh/AFh): SSD1305 supports dim mode, whereas SSD1306 only has the standard on/off mode.
Set Display Clock Divide Ratio/Oscillator Frequency (D5h): SSD1305 and SSD1306 have different default values and adjustment ranges.

Commands exclusive to SSD1306:

Charge Pump Setting (8Dh): Manages the built-in charge pump for display power supply.
These commands and their differences reflect the main distinctions in functionality and control between SSD1305 and SSD1306. SSD1305 offers more options for color and brightness control, while SSD1306 has simplified commands for monochrome display and additional features like fade-out and blinking modes._

I haven't studied your code completely, but I don't think you're using Fade Out, Blinking, and Zoom. Display on/off works fine. Tested on MT32pi. Main difference - 1305 grayscale/color. But now in markets not present color models.

You can, of course, remove the code modifications for the 1305, but then many projects won't be able to use the 1305 instead of the 1306. Perhaps it would be better to keep these changes for monochrome use of the 1305, while for full-color or grayscale configurations, use a different driver—which, by the way, might never be written or is uncertain when it will be available.

rsta2:
I was hoping, someone would be able to port and test the SSD1305 support from the class `CSSD1306Device` to the new class `CSSD1306Display`, which provides a more comprehensive display support for the SSD1306.

PR #630, opened 2025-12-07
Fix cache synchronize not clearing full length

makuto:
If the address is not cache-line aligned, then the length it is not aligned will not be properly cleared, because each iteration assumes an entire cache line's worth has been cleared.

Instead, we first align to the next smallest cache line and add the difference so we are sure to clear at least the full length requested.


rsta2:
This makes sense. I think luckily there was not a real problem with this, because DMA buffers have to be cache-aligned in Circle, but it's better to do it this way. Thank you for the PR!

PR #631, opened 2025-12-11
Initialization-time VID/PID logic for USB Gadgets

danifunker:
Initialization-time VID/PID logic for usbmidigadget, usbcdcgadget and usbmsdgadget


rsta2:
Thank you for the PR!

PR #638, opened 2026-01-09
Support configurable encoder types in KY040 driver

speendo:
### Summary
**This PR adds support for quarter-step and half-step rotary encoders to the KY040 driver, while maintaining backward compatibility with full-step encoders (default behavior).**

Different rotary encoder hardware uses different detent mechanisms. The existing implementation only worked correctly with full-step encoders, making it incompatible with other variants.

### Changes

* Added nEncoderDetents parameter to CKY040 constructor (values: 1, 2, or 4; default: 4 for full-step)
* Replaced complex state machine with simpler lookup table approach
* Added step counter that accumulates transitions until threshold is reached
* Events are fired only when |counter| >= detents, providing consistent behavior across encoder types
* Handles contact bounce gracefully (invalid transitions return 0 direction)

### Technical Details

The new implementation uses a 16-entry lookup table that maps all possible pin state transitions to rotation direction (-1, 0, or +1). The step counter accumulates these values, and when the threshold is reached, an event is fired. This approach is simpler and more robust than the previous state machine. Invalid transitions (both pins change at the same time) also result in 0 (= no step count).

This provides direction detection while tolerating the imperfect Gray code sequences that real-world encoders produce due to contact bounce or missed steps.

### Testing

Tested with a half step encoder on a minidexed build with a Raspberry Pi 3:
detents=2 results in the expected behaviour (each detent is registered once)
detents=1 results in two steps per detent
detents=4 results in the previous behaviour (only every second detent is registered).

It's also worth noting, that previously the encoder was only responding after a start state is reached. Now the Encoder reacts right after system start.

Memory impact: Adds ~ 4 bytes per encoder instanct

### DIFF
--- a/addon/sensor/ky040.cpp
+++ b/addon/sensor/ky040.cpp
@@ -23,38 +23,27 @@
 static const unsigned SwitchDebounceDelayMillis	= 50;
 static const unsigned SwitchTickDelayMillis = 500;
 
-CKY040::TState CKY040::s_NextState[StateUnknown][2][2] =
+// Encoder direction lookup table
+// Indexed by 4-bit pattern: (oldCLK << 3) | (oldDT << 2) | (newCLK << 1) | (newDT << 0)
+// Returns: -1 for counter-clockwise, 0 for invalid/no-change, +1 for clockwise
+static const s8 s_EncoderDirection[16] = 
 {
-	// {{CLK=0/DT=0, CLK=0/DT=1}, {CLK=1/DT=0, CLK=1/DT=1}}
-
-	{{StateInvalid,    StateCWStart},      {StateCCWStart,    StateStart}},   // StateStart
-
-	{{StateCWBothLow,  StateCWStart},      {StateInvalid,     StateStart}},   // StateCWStart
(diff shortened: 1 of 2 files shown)


soyersoyer:
I tested it with a full and a half encoder. They work well and are now precise without the [suggested](https://github.com/probonopd/MiniDexed/wiki/Hardware#rotary-encoder) capacitors.
Thank you.

One little thing, I think you should follow the file's indentation style. ([Allman](https://en.wikipedia.org/wiki/Indentation_style#Allman_style))

rsta2:
@speendo Thank you for this improvement of the driver!

@soyersoyer Thanks for testing!

speendo:
Thanks for testing @soyersoyer! I created another PR (#639) to fix the indentation.

The improved precision is a welcome side effect. I think it comes from the fact that contact bounce is handled gracefully - both directions of a bounce are counted (e.g., bounce up: +1, bounce down: -1), which leads to a neutral result in the step counter.

Thank you @rsta2 for the fast response and accepting my PR!

PR #639, opened 2026-01-11
Fix indentation to match Allman style

speendo:
All functionality unchanged, formatting only

Addresses feedback from @soyersoyer in https://github.com/rsta2/circle/pull/638#issuecomment-3731298480


rsta2:
Thanks!

PR #649, opened 2026-02-16
Add SSD1309 OLED display driver (I2C + SPI)

Cinezaster:
Add CSSD1309Display graphics-mode driver for the SSD1309 128x64 OLED display controller. The SSD1309 is register-compatible with the SSD1306 but requires a mandatory hardware reset pin and uses external VBAT (no internal charge pump). Key initialization differences:
- No charge pump command
- Pre-charge period 0xF1 (vs SSD1306's 0x22)
- VCOMH deselect level 0x34 (vs SSD1306's 0x20)

The driver supports both I2C and SPI interfaces via separate constructors, following the dual-transport pattern used by CILI9341Display for SPI.

Includes:
- addon/display/ssd1309display.h and .cpp
- SPI and I2C configuration blocks in sampleconfig.h
- Sample program in addon/display/sample/ssd1309/

Tested on Raspberry Pi 3 (AArch64) with 2.4" 128x64 SSD1309 OLED via both I2C (0x3C) and SPI (8 MHz, CPOL/CPHA 0/0).

### DIFF
--- a/addon/display/Makefile
+++ b/addon/display/Makefile
@@ -5,7 +5,7 @@
 CIRCLEHOME = ../..
 
 OBJS	= hd44780device.o st7789display.o ili9341display.o ssd1306display.o \
-	  chardevice.o ssd1306device.o st7789device.o
+	  ssd1309display.o chardevice.o ssd1306device.o st7789device.o
 
 libdisplay.a: $(OBJS)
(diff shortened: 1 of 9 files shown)


rsta2:
Thank you for this contribution and the comprehensive info!

PR #659, opened 2026-03-17
add memem to utils

alexcb:
This adds a memem function, similar to strstr, but for data.


rsta2:
Thanks for this contribution!

PR #662, opened 2026-04-06
Implement container sanitizer helpers for libc++

smuehlst:
When Kernel Address Sanitizer is enabled with the new libc++ builds, additional runtime support is necessary to instrument container classes.

Generated-by: Claude Code Pro


rsta2:

PR #677, opened 2026-06-29
Add sh1106 oled support.

penfold42:
NOT my work, I just extracted it from
https://github.com/probonopd/MiniDexed/pull/1011

I have tested it against:
- the circle commit that minidexed is using.
- this branch now that I've added a sample

### DIFF
--- a/addon/display/Makefile
+++ b/addon/display/Makefile
@@ -5,7 +5,7 @@
 CIRCLEHOME = ../..
 
 OBJS	= hd44780device.o st7789display.o ili9341display.o ssd1306display.o \
-	  ssd1309display.o chardevice.o ssd1306device.o st7789device.o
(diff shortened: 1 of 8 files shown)


rsta2:
The class `CCharDevice` is deprecated in Circle. New display driver classes have to be derived from `CDisplay` and must use `CTerminal` for character display. That's why I cannot merge this.

penfold42:
ok.

is their support in the CTerminal/CDevice framework for hd44780 style displays?

rsta2:
Good point! No, it's only for dot-matrix displays. So my statement was 
wrong. It must me: The C*Device classes for dot-matrix displays in 
addon/display/ are deprecated and the class CTerminalDevice should be 
used instead. CCharDevice must remain for CHD44780Device, which uses it 
for its implementation. Thanks for asking!

PR #682, opened 2026-07-23
https://github.com/rsta2/circle/compare/master...HERMITRION:circle:fix/usb-cold-boot-hang

HERMITRION:
## Summary

Fixes a reliably-reproducible full system hang on the DWHCI USB host driver
when a USB device (e.g. a USB MIDI keyboard) is already attached at cold boot.

This was originally reported and diagnosed in
[probonopd/MiniDexed#1042](https://github.com/probonopd/MiniDexed/issues/1042).

## Root cause

When a USB device is already attached at cold boot, the port reset sequence
in `CDWHCIDevice::EnableRootPort()` reliably (100% reproducible on a
Raspberry Pi Zero W) triggers a spurious "disconnect" event for the
still-attached device. This event is queued and processed asynchronously via
`CUSBHostController::UpdatePlugAndPlay()`.

If a blocking control transfer (part of the normal enumeration /
`CUSBDevice::Configure()` sequence) happens to be in flight when this
spurious disconnect is processed, `CDWHCIRootPort::RemoveDevice()`
immediately deletes the `CUSBDevice` object without releasing the DWC2
channel that transfer was using. Since control transfers use
`USB_TIMEOUT_NONE` and `CDWHCIDevice::TransferStage()` busy-waits on
`m_bWaiting[nWaitBlock]` with no timeout, the channel interrupt that would
normally clear this wait never arrives (the device object it belonged to is
already gone) — so the calling task spins forever. On a single-core /
cooperative-scheduler setup this hangs the whole system.

This also explains why hot-plugging after boot is unaffected: by the time
the device is plugged in, the port reset from `EnableRootPort()` has already
happened, so there is no reset-induced spurious disconnect to race against.

## Fix

`CDWHCIRootPort::RemoveDevice()` now explicitly aborts any channel still
active for the device being removed, via a new
`CDWHCIDevice::AbortActiveChannels()` called from
`CDWHCIDevice::CancelDeviceTransactions()`, *before* the `CUSBDevice` is
deleted.

`AbortActiveChannels()` disables the channel through the normal DWC2
sequence (clear `DWHCI_HOST_CHAN_CHARACTER_ENABLE`, set
`DWHCI_HOST_CHAN_CHARACTER_DISABLE`, unmask `DWHCI_HOST_CHAN_INT_HALTED`).
This lets the existing `!m_bRootPortEnabled` cleanup path already present at
the top of `ChannelInterruptHandler()` complete the pending URB with
`USBErrorAborted` and free the channel, instead of leaving the caller
spinning forever.

This does not prevent the spurious disconnect event itself (which appears to
be an inherent characteristic of the DWC2 port reset sequence on this
hardware), but it makes the driver resilient to it: the device is safely
torn down and re-enumeration proceeds normally on the next
`UpdatePlugAndPlay()` cycle instead of freezing the system.

## Changes

- `include/circle/usb/dwhcidevice.h`: declare `AbortActiveChannels()`
- `lib/usb/dwhcidevice.cpp`: implement `AbortActiveChannels()`; call it from
  `CancelDeviceTransactions()`
- `lib/usb/dwhcirootport.cpp`: call `CancelDeviceTransactions()` in
  `RemoveDevice()` before deleting the device

## Testing

- Reproduced 100% of the time on a Raspberry Pi Zero W with a USB MIDI
  keyboard attached at cold boot, prior to this fix.
- After the fix, cold boot with the device attached completes normally, the
  device is enumerated, and normal operation (MIDI in this case) proceeds —
  no freeze.
- Tested with KORG nanoKey.
- Only tested on Raspberry Pi Zero W (single core, non-SOF-interrupt build).
  I have not been able to test this on multi-core builds or with
  `USE_USB_SOF_INTR` enabled — the logic should apply equally there (channel
  allocation only happens once a transaction actually starts on a channel,
  regardless of `USE_USB_SOF_INTR`), but additional testing on those
  configurations would be welcome.

This is a fix for the *hang*; it does not address the root cause of why the
spurious disconnect event is generated in the first place, which appears to
be inherent to the port reset timing on this controller.


rsta2:
I tried to replicate the reported issue with a RPi Zero W and an AKAI LPK25 USB MIDI keyboard controller, but the problem does not happen there. Nevertheless it may happen on some other controllers. 

Unfortunately there is a problem in your fix. When using `MSEC2HZ()`, you have to use `CTimer::GetTicks()` (returns `HZ` units) instead of `CTimer::GetClockTicks()` (returns microseconds). Before your PR can be accepted, please also replace the Japanese comments with English ones (without the "NEW:" prefix). As mentioned in *CONTRIBUTING.md*, Circle uses tabs, which are 8 characters wide. This should be the case in your modifications too. Thanks so far.

This patch must be tested very well, because it touches important parts of the DWHCI USB driver.

HERMITRION:
Thanks for catching that — fixed:

- Switched `GetClockTicks()` → `GetTicks()` (paired correctly with `MSEC2HZ()`) in both `BeginRecoveryBackoff()` and `UpdatePlugAndPlay()`.
- Replaced all Japanese comments with English, removed the `NEW:` prefixes.
- Converted the affected blocks to tab indentation (8-wide) per CONTRIBUTING.md.

I made these changes without my test hardware on hand, so I haven't been
able to re-verify on the Pi Zero W yet. The logic itself is unchanged (just
correctly implementing the same intent as before), but since you mentioned
this needs thorough testing, I'll hold off on saying it's fully verified
until I'm back at my setup and can confirm cold boot with the device
attached still behaves correctly.

rsta2:
Thanks for the changes! I will merge it, when you have tested it again.

HERMITRION:
Update: I've now been able to re-test on hardware with the review-fixes
commit applied (`GetTicks()` fix, English comments, tab indentation).

Cold boot with a USB MIDI keyboard already attached completes cleanly on my
Raspberry Pi Zero W — no freeze, device enumerates normally, and MIDI/audio
work as expected. Confirmed this holds up after the changes, not just with
the original patch.

rsta2:
Thanks for testing! Let me also test this a little bit. I will merge 
your PR in the next two days.

rsta2:
I tested your patch to some degree and need some more modifications:

* When the `override` keyword is added to a virtual method, all overridden methods must have this keyword in a class declaration, otherwise Clang shows a warning. That's the case with `UpdatePlugAndPlay()`. I would simply remove this keyword there.

* These member initializers are not in the right place, they must before `m_pInterruptSystem`, otherwise a warning is shown:


```
	m_nRecoveryAttempts (0),
	m_nRecoveryRetryAtTicks (0),
	m_bRecoveryGiveUp (FALSE),
```

* The comment about the initializers is not necessary in *dwhcidevice.cpp* line 52.

* Circle is normally relatively quiet, when no error occurs. I don't want to have a long bunch of messages, when the system boots. That's why please remove all log messages from your patch. There should be only one warning "No device connected to root port" (without attempt number), when there is no USB device at all attached (as before). BTW. This is not a serious error, because there are applications, which work without an attached device, even when an USB host controller driver is installed (like MiniDexed with serial MIDI).

rsta2:
Because I wanted to merge this today, I made the requested changes on my own now. Because there is a new log level `LogTrace` in Circle now, I did not remove your boot recovery messages, but replaced them with `LOGTRACE()`. They can be enabled with `loglevel=5` in the file *cmdline.txt* on the SD card.

I picked your commits manually, because there were commits from the *master* branch in this PR, which I didn't want to have on *develop* with this merge. You will be listed under the Contributors, when the next release is done.

Thank you for researching this issue and for providing the fix!

HERMITRION:
Thank you for reviewing this so thoroughly and for taking the time to
integrate it properly — really appreciate it, and glad the `LOGTRACE()`
approach found a good home for those messages too. Happy to have helped!

Issue #531, opened 2025-01-06
Specific Version of The64Maxi Keyboard doesn't work in Circle 48


Chris-Hal:
Hello Rene, 
I would like report one issue with the Circle library in combination with Randy Rossi BMC64 project.
https://github.com/randyrossi/
To be honest, Randy suggested to me, to open an issue, because I was already in contact and our findings seems to be an incompatible USB device in combination with the circle library. Additionally there are already users, which are using his BMC64 on Raspberry Pi 2/3   together with a modified TheC64 Maxi hardware.
https://retrogames.biz/products/thec64/

From the TheC64 Maxi only the USB keyboard will be used to the Pi and the internal USB hub of the Maxi, so all together becomes a perfect replica of an C64. So did I with an RPI2 B version. Sorry if I bother you wiith such details in case you know this already. 

When I stick "TheC64" Keyboard to the Raspberry Pi and start his BMC64 which comes Originally with circle 40 the Pi will stuck as visible in this screen.
![Maxi_keyb_connected_IMG20250106092929_lq](https://github.com/user-attachments/assets/a09ef965-b9a9-431e-a37d-0ebbd25aa211)

If  I connect no keyboard at all to the Raspberry Pi, than BMC with Circle boots normally but without Keyboard input.
![maxi_keyb_disconnected_IMG20250106093031_lq](https://github.com/user-attachments/assets/a70be4e1-a59e-47ea-abbc-5a82b55bfc85)

The 3rd option was to connect any noname Wireless Keyboard dongle to the Raspberry which looks like this and typing is possible.
![USB Dongle_IMG20250106125053_lq](https://github.com/user-attachments/assets/c0eb8442-cf01-41e0-8df9-f0520020d878)

Randy created for me a test enviroment with Circle library 48 but even here the boot process stuck when detecting the TheC64 keyboard.
![Circle48_Maxi_KB_connected_IMG20250106145124](https://github.com/user-attachments/assets/b7672386-75be-49e5-aa0f-51fd39d1b374)

[...]

PR #688, opened 2026-08-06
Fix TheC64 keyboard handling for the two versions

aminch:
Distinguish the two THEC64 keyboard firmware versions by HID report descriptor length. Only the 78-byte descriptor requires the additional report byte; the other version keeps the standard report size.

This was a problem that was reported in the BMC64 Github repo here: https://github.com/randyrossi/bmc64/issues/290 where it was discovered that there is a different revision of THEC64 keyboard that didn't work with the original patch (https://github.com/rsta2/circle/commit/4de2b0970e9ebe47987a282d17b6aa51e487c446) even though it as the same product and vendor id.

Details of my investigation and follow up are in the orignal bug reported to Circle where I contacted the orignal reporter and tested this patch with both versions of THEC64 that have been discovered. See: https://github.com/rsta2/circle/issues/531#issuecomment-5167878893 

Both parties (@Chris-Hal and @bugjacobs) reported that this fix works.


rsta2:
Thank you for this fix, also to the testers!

Issue #681, opened 2026-07-13
[Bug] Socket->Send() Sometimes Corrupted Data / Resending Previously Sent Data / Slow Chunking


davefilip:
> You have found a bug in Circle? Please fill out this form!

**Description of the bug**

Rene,

I have recently started upgrading my Circle-based OS from Step 50 to Step 51, given that it sounds like the network layer has been improved, and I am excited to try the new IPC (pipe) functionality.

The upgrade was easy, and I didn't need to modify any code this time (vs. going from 49 to 50 which required some refactoring).

However, I did start having a problem with Step 51 related to network sockets.

I have a telnet server that I wrote 2+ years ago, that had been working reliably since.  However, with Step 51 I've been having some reliability problems, which seems to be related to calling CSocket->Send():

	CLogger::Get()->Write(LogSource, LogNotice, "[%d] %s", strlen(string), string);
	m_Socket->Send(string, strlen(string), MSG_DONTWAIT);

I added the Clogger line before it to verify that the string being passed is good, and the length is valid (i.e., not is never an unterminated string, the parameters are valid as displayed on the conosle).

Although I have no problem receiving data from a telnet client, sending data to the telnet client with the same kernel may or may not work depending on the RPi model and WiFi access point.

Sometimes it works fine, but sometimes will re-send fragements of data that has previously been sent, or add what looks like lots of blank lines, and sometimes it also displays the proper text but gets very choppy (seconds between each fragment of text sent).

I have done a lot of testing, and I suspect that it might be related to some combination of processor and network speed, based on the following (all using the EXACT SAME kernel image):

A. RPi 3B + High Speed Mesh WiFi - have not seen any problems

B. RPi 3B + Medium Speed Pocket WiFi (1) - mostly reliable, can occasionally see a problem

C. RPi 3B + Low Speed RPi Hotspot (2) - Frequently works but can sometime see the problem

D. RPi Zero 2W + High Speed Mesh WiFi - Frequently works but can sometime see the problem

E. RPi Zero 2W + Medium Speed Pocket WiFi (1) - Sometimes works but frequent problems

F. RPi Zero 2W + Low Speed RPi Hotspot (2) - Lots of problems, mostly unusable

Note: 1 - MikroTik mAP (2x2 inch battery)
Note: 2 - Raspberry Pi 3B running HostAPD + DNSMasq

For my current project, ideally I need to get (F) working for an IoT project on a remote platform.

[...]

PR #689, opened 2026-08-06
Fix Circle TCP ACK handling for fragmented modem sessions

aminch:
I discovered this problem when implementing a software modem to use with the BMC64 project. My original commit in my fork of that project is here: https://github.com/aminch/bmc64/commit/af27ebac49ba21145fde4fe775bea498882835d3

I have only tested this on the BMC64 builds I have made. If there additional testing or test case I should run please let me know.

### Reasoning for fix

Circle left acknowledged one-byte application sends in its transmit queue while established. CCGMS commonly sends single-byte keystrokes, so later incoming ACKs could make fast retransmit send an already acknowledged byte at an old sequence number. The peer then considered subsequent fragmented data out of order, causing the session to stall or disconnect.

Flush every acknowledged application byte from the transmit queue, including one-byte sends. Restrict duplicate-ACK processing to RFC 5681 conditions: the ACK must equal SND.UNA, data must be outstanding, the segment must be ACK-only, and its advertised window must be unchanged. Reset the duplicate-ACK count for all other ACKs

### DIFF
--- a/lib/net/tcpconnection.cpp
+++ b/lib/net/tcpconnection.cpp
@@ -1230,12 +1230,6 @@ int CTCPConnection::PacketReceived (CNetBuffer	*pPacket,
 					m_bFINQueued = FALSE;
 				}
 
-				if (   m_State == TCPStateEstablished
-				    && nBytesAck == 1)
-				{
-					nBytesAck--;
-				}
-				
 				if (nBytesAck > 0)
 				{
 					m_TxQueue.Flush (nBytesAck);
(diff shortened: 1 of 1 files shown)


rsta2:
Thank you for this fix! I need to test it first, before merging it.

rsta2:
Thanks again for this fix! I tested it successfully with the samples 20, 21, 35 and with the iperf2-test in [Unitest](https://github.com/rsta2/unitest) (client and server mode).

Unfortunately there aren't much network (especially TCP) tests for Circle. If somebody has an idea, which TCP stress test for Linux could be ported to Circle, let me know! Also with issue #681 in mind, this would help a lot. With the POSIX-network-support in circle-stdlib porting could be not a great effort.

PR #690, opened 2026-08-13
Fix synchronizationevent lost wakeup

Valefungo:
Hello René,

I'm making a version of 86Box that can run on baremetal. That project is an heavy multi core and multi-CTask thing and exposed a real problem because it has a concurrent emulation on core 1 and seven CTask on core 0.

Wait()/WaitWithTimeout() check m_bState, then call BlockTask() to actually block - two separate steps, not atomic. Under ARM_ALLOW_MULTI_CORE, another core's Set() can land in the gap between them: it sees the wait list still empty, wakes nobody, and marks itself already-fired (so a later Set() is a no-op) - while the waiter, having already decided to block from the now-stale check, blocks anyway right after with nothing left to ever wake it.

Found this via GDB on a real hang: a worker CTask stuck Blocked forever on an event that was already TRUE, with real pending work behind it nobody would process.

Fix: BlockTask() takes an optional state pointer and re-checks it itself, under the same lock Set()/WakeTasks() use so the check and the registration become one atomic step instead of two.

Tested on a Pi Zero 2, I'll test this on a Pi 4 soon

### DIFF
--- a/include/circle/sched/scheduler.h
+++ b/include/circle/sched/scheduler.h
@@ -111,7 +111,19 @@ private:
 	void AddTask (CTask *pTask);
 	friend class CTask;
 
-	boolean BlockTask (CTask **ppWaitListHead, unsigned nMicroSeconds);
+	// aarch64pi note (2026-08-12): pState (optional, default null) closes
+	// a real lost-wakeup race - see the implementation in scheduler.cpp
+	// for the full story (confirmed via GDB on a real hang: an S3 ViRGE
+	// FIFO worker CTask left permanently Blocked on an event whose
+	// m_bState was already TRUE, because core 1's Set() ran - and found
+	// the wait list still empty - in the gap between the caller's own
+	// state check and this function's wait-list registration below,
+	// which used to be unprotected by any lock). When pState is given,
+	// *pState is re-checked under the same m_SpinLock WakeTasks()/Set()
+	// use, atomically with the registration - if already true, this
+	// returns immediately instead of blocking at all.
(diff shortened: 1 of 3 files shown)


rsta2:
Thanks, but it's a Circle requirement, that the scheduler and all task stuff runs on core 0. That does also mean, that one cannot manipulate a synchronization object (e.g. event, semaphore) from a secondary core. Even if it works for you with this patch, there may be other issues with it. To relax this (core 0) requirement would need more work. I'm currently not ready to do this.

Valefungo:
Hello Rene,

WakeTasks() is documented as callable from interrupt context, so this race exists on a single core too: an ISR calling Set() can land right between Wait()'s old state check and its BlockTask() call, find the wait list still empty, and never wake anyone - no second core needed.
The fix just moves that check inside the same critical section m_SpinLock already uses.

Anyway it's up to you, I'm keeping this in my codebase othewise it dies.

Kind Regards,
Valerio

rsta2:
I have to check this. Will take some time.

rsta2:
Hi Valerio, OK, there can be a race condition, when Set() is called from an IRQ handler, before the blocking operation in Wait() takes effect.

Your patch has been applied, but with reduced comments, because the fix has been already described sufficiently in your commit notice, and manually, because you specified the *master* branch, when branching off and as the base branch for merging. The *develop* branch is required here (see the *CONTRIBUTING.md* file).

Thank you for this contribution! You will be listed as contributor, when the patch has been merged to the *master* branch with the next release.

PR #691, opened 2026-08-14
Don't use blocking USB bulk-IN receives when ethernet link is down

aminch:
Another fix while developing BMC64.

Fix emulator BMC64 slow downs while Ethernet link is down

BMC64 starts networking asynchronously so emulation can launch before Ethernet has a physical link or DHCP address. Circle then runs its network task in the background and repeatedly polls the selected network device.

The LAN7800 and SMSC951x drivers used blocking USB bulk-IN receives even when the PHY reported no link. With the cable unplugged, this could leave the network task waiting indefinitely and cause severe emulation, video, and audio slowdowns.

Fixed by tracking PHY link state through Circle's periodic PHY task and skip receive submissions while the link is down. This covers SMSC951x on Pi 2/Pi 3B and LAN7800 on Pi 3B+.

The original patch in BMC64 can be see here: https://github.com/randyrossi/bmc64/commit/c506c424692c0b104d290de568c39ad95c2a1ad2

FYI: @randyrossi here is the ethernet patch submission.

### DIFF
--- a/include/circle/usb/lan7800.h
+++ b/include/circle/usb/lan7800.h
@@ -45,6 +45,7 @@ public:
 
 	// returns TRUE if PHY link is up
 	boolean IsLinkUp (void);
+	boolean UpdatePHY (void);
 	
 	TNetDeviceSpeed GetLinkSpeed (void);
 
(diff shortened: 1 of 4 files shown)


rsta2:
This makes sense. Thank you for this contribution!

BTW. Please use a maximum line length of 72 characters for creating commit notices for further PRs, because the notices are difficult to read otherwise in a terminal window. This line length is a common standard.

aminch:
Thanks, I'll check the comment length next time

PR #692, opened 2026-08-15
circle-stdlib returns to GitHub

smuehlst:
Restored the GitHub links for circle-stdlib, as it migrates from Codeberg back to GitHub because of Codeberg's new policy regarding software developed with AI assistance.


rsta2:
Thank you for the PR!

PR #696, opened 2026-08-30
Fix Wifi unable to connect to eero7 router

aminch:
Association request failed with "bad wpa ie syntax" due to an optional IE trailing the RSN/WPA2 IE. Optional IE now ignored. 

Detailed analysis here: https://github.com/randyrossi/bmc64/issues/335

The user with the problem confirmed it is fixed, and existing users seem unaffected.  I only have so many Wi-Fi routers I can test.


rsta2:
Thank you for researching the problem and for providing the fix! I successfully tested it on my local AP, also with hidden SSID.

PR #685, opened 2026-07-29
8bitdo xbox360 keyboard and gamepad support

aminch:
- Support for 8bitdo Retro Keyboard
- Support for various 8bitdo controllers
  - Pro 3 controller needed the additional: "Set port configured after the configure call"
  - M30 2.4G controller needed: "For Pi4 pr greater devices read the entire descriptor at once."
- Support for Xbox 360 PC Wireless Receiver (and clone)

Please let me know if you need further explainations, I hope there is enough information in the code comments and commits.

PR #700, opened 2026-09-12
usbkeyboard8bitdo: Fix LED status reports

aminch:
CUSBKeyboard8BitDoDevice did not override SetLEDs(), so the base
class implementation was always used, sending LED output reports
with report ID 0. This device expects report ID 1, so caps lock,
num lock and scroll lock LEDs never illuminated.

Add the SetLEDs() override that prefixes the report with ID 1,
and mark CUSBKeyboardDevice::SetLEDs() virtual so it dispatches
correctly.

Small addtion to my original pull request: https://github.com/rsta2/circle/pull/685


rsta2:
Thank you!

PR #701, opened 2026-09-12
Fix out-of-bounds write on DWHCI exhaustion

aminch:
Found while debugging a composite device with 6 concurrently polled HID interfaces. SOFInterruptHandler() used an out-of-range channel index when none were free; the guarding assert() is compiled out in release builds, so it silently corrupted memory past the end of m_pStageData[] instead of failing safely. Re-queue the transaction for the next frame instead.

This was found when attempting to fix support for Keyrah in BMC64: https://github.com/randyrossi/bmc64/issues/235#issuecomment-5646932517

I am not sure what other specific testing should be done to check this further. Please let me know if you need more details.


rsta2:
I never saw this assertion failed, but it could happen with many devices attached. The solution looks fine. Thank you!

PR #704, opened 2026-09-17
Recover xHCI endpoint from Halted state on transfer error

aminch:
While testing keyboards on Pi4B to find the issue for #685 I discovered this issue when connecting 8bitdo Retro Keyboard in wired mode.

CUSBEndpoint::ResetPID() was a no-op on Pi 4/5 (xHCI), so a single Stall or transfer error on any control/bulk endpoint, including EP0, left the xHC's endpoint context Halted with no recovery. Every later transfer on that endpoint then silently timed out instead of completing, which showed up as enumeration failing outright for some USB devices (e.g. the 8BitDo Retro Keyboard in wired mode: one stalled control transfer during setup, then SetConfiguration timed out and the device was dropped).

When creating a Pull Request, please be sure to select the *develop* branch as base, where your PR should be merged!


rsta2:
Thank you!

